PR #117, opened 2014-04-09
Add a TEE modifier.

eevee:
I happened to need this recently, when wrapping a program that writes things I need to know to `stdout`, but occasionally prompts for user credentials on `stdout` as well.  It's pretty dumb, but seems to do the job.

### DIFF
--- a/plumbum/commands/modifiers.py
+++ b/plumbum/commands/modifiers.py
@@ -1,3 +1,8 @@
+import os
+from select import select
+from subprocess import PIPE
+import sys
+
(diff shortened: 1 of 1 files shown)


gaschreiber:
It would be great to make the TEE available where FG and BG are available too.
plumbum/commands/__init__.py:2 reads:
from plumbum.commands.modifiers import ExecutionModifier, Future, FG, BG
The TEE should be added to this list?

DanielJoyce:
Yes it should be added to that list. Its goofy right now

PR #127, opened 2014-06-09
Various small improvements/fixes

chaoflow:


henryiii:
By the way, [colored](https://github.com/dslackw/colored) is a nice library for colors in Python. I would love to have something built in, though, especially with context manager support, like I use [here](http://iscinumpy.blogspot.com/2015/07/plumbum-scripting.html#Putting-it-together-in-an-example:-git-all).

Issue #140, opened 2014-08-01
Add root app reference for subcommands.


khorn:
Since `cli.Application` objects can be set up in arbitrarily deep trees, it would be convenient to have a way to refer to the "root" `cli.Application` instance from anywhere in the tree.

PR #141, opened 2014-08-01
Add a root_app property to `cli.Application`

khorn:
This adds a property to `cli.Application` called root_app which returns a
reference to the "root" Application object in a tree of Application objects.

Basic implementation for issue #140.


metaperl:
Can this be documented in the User Guide?

Issue #173, opened 2015-02-07
Windows support for ssh with password


fruch:
I was trying to do this:

    mach = SshMachine("10.57.33.XX", user="root", password="something")

and it failed on:

    plumbum.commands.processes.CommandNotFound: ('sshpass', ....

looking at the code, I've found PuttyMachine... that should work better...

    mach = PuttyMachine("10.57.33.XX", user="root", password="something") 
    # yielded this 
    Traceback (most recent call last):
    File "C:/split_braaa.py", line 10, in <module>
        mach = PuttyMachine("10.57.33.XX", user="root", password="something")
    TypeError: __init__() got an unexpected keyword argument 'password'

So how can I use Plumbum SSH on windows ?

fruch:
Later I've found out the ParamikoMachine, but not much luck with getting it to work on Windows.
together with rpyc ZeroDeply feature, I was just hanging there when calling `server.classic_connect()`

the PuttyMachine was working at the end, but with breaking linux support (both in rpyc and plumbum, as can seen in their TravisCI pages )

msvbhat:
Hi,

I tried connecting to Windows machine from linux client. I was running freesshd in windows to get access to powershell. And from the bash, I can login to powershell (just like ssh) and run my commands.

But when I tried to connect via plumbum.SshMachine (from linux to windows) I get the backtrace.

>>> from plumbum import SshMachine
>>> rem = SshMachine("xx.xx.xx.xx", user="root")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.6/site-packages/plumbum/machines/ssh_machine.py", line 106, in __init__
    new_session = new_session)
  File "/usr/lib/python2.6/site-packages/plumbum/machines/remote.py", line 129, in __init__
    self._session = self.session(new_session = new_session)
  File "/usr/lib/python2.6/site-packages/plumbum/machines/ssh_machine.py", line 150, in session
    self.encoding, isatty, self.connect_timeout)
  File "/usr/lib/python2.6/site-packages/plumbum/machines/session.py", line 141, in __init__
    self.run("")
  File "/usr/lib/python2.6/site-packages/plumbum/machines/session.py", line 224, in run
    return run_proc(self.popen(cmd), retcode)
  File "/usr/lib/python2.6/site-packages/plumbum/commands/processes.py", line 186, in run_proc
    stdout, stderr = proc.communicate()
[...]

PR #174, opened 2015-02-07
add password support for PuttyMachine()

fruch:
- add password support for PuttyMachine()
- fix .tunnel() to work also work in windows

rpyc ZeroDeploy wasn't working that good in windows, and hence those fixes

related to:
tomerfiliba/plumbum#173


tomerfiliba:
that's windows specific. openssh doesn't have ``-pw`` switch, and the hostname must be bracketed if  ipv6 is used, since it uses colons (e.g., ``[::1]:22``)

Issue #178, opened 2015-02-08
`env` only works on linux-like systems


abarnert:
The fix for #162 uses `env -0`, which is a GNU extension that isn't part of POSIX, BSD, etc., which means a whole lot of things now break on OS X, FreeBSD, etc.

As far as I know, there is no perfect portable fix for #162; for example, the standard `env` command will produce the exact same output for these two cases:

    env a=foo b=bar
    env a='foo
    b=bar'

As far as I know, the only way to do it reliably and portably (at least to anything that's compatible with BSD3 or later, which is a lot more than just GNU…) from the shell is to first get a (potential) list of names by calling and parsing `env` or `printenv`, and then lookup each name one by one with `printenv`.

Of course that can obviously cause performance problems. You can solve that for GNU systems by trying `env -0` first and only falling back if that fails:

        env0 = session.run("env -0; echo")
        if env0[0] == 0 and not env 0[2].rstrip():
            self._curr = dict(line.split('=', 1)
                              for line in env0[1].split('\x00')
                              if '=' in line)
        else:
            lines = self.remote._session.run("env; echo")[1].splitlines()
            split = (line.split('=', 1) for line in lines)
            keys = (line[0] for line in split if len(line)>1)
            runs = ((key, session.run('printenv "%s"; echo' % key))
                    for key in keys)
            self._curr = dict((key, run[1].rstrip('\n')) for (key, run) in runs
                              if run[0] == 0 and run[1].rstrip('\n')
                              and not run[2])

Alternatively, for platforms without `env -0`, you can use some heuristics on the output of `env`, which aren't perfect, but may be good enough for most users.

Your existing code already skips any line without a `=`; this will truncate variables with embedded newlines, and it will mean that `a='foo\nb=bar'` will give you a fake variable `b`, but maybe that's acceptable.

You can eliminate the first problem by treating any line that doesn't start with a valid name followed by `=` as a continuation of the previous line, as in pigay's suggested patch for #162.

But there's no way to eliminate the last problem. I don't *think* that's likely to fail too often by accident… but you might want to think about whether someone could engineer some exploit that takes advantage of it.

henryiii:
I believe this was fixed with #179. Closing.

Issue #162, opened 2014-11-02
Strange remote environment variable causes error


pigay:
Hi,

with a SshMachine, strange remote environment variable causes an error:

```
Traceback (most recent call last):
  File "rpyc_test.py", line 5, in <module>
    mach = SshMachine("avakas.mcia.univ-bordeaux.fr", user="pigay", keyfile="/home/pigay/.ssh/id_rsa")
  File "/usr/local/lib/python2.7/dist-packages/plumbum/machines/ssh_machine.py", line 106, in __init__
    new_session = new_session)
  File "/usr/local/lib/python2.7/dist-packages/plumbum/machines/remote.py", line 132, in __init__
    self.env = RemoteEnv(self)
  File "/usr/local/lib/python2.7/dist-packages/plumbum/machines/remote.py", line 18, in __init__
    self._curr = dict(line.split("=", 1) for line in self.remote._session.run("env")[1].splitlines())
ValueError: dictionary update sequence element #44 has length 1; 2 is required
```

The environment that causes this error is the second part of a function definition. Here is what appears with an `env` command on the remote machine:
```
BASH_FUNC_module()=() {  eval `/usr/bin/modulecmd bash $*`
}
```

The return character in BASH_FUNC_module causes an error in `machines/remote.py` because of splitline call, I think.

Would it be ok to simply catch the ValuerError, or to use a more complicated parsing method?

Regards,

Pierre

campbellr:
I've also been hitting this.

I have the following environment variable that breaks due to `splitlines`: `'COMP_WORDBREAKS= \t\n"\'><;|&(:'`

This seems to fix it:

```
- self._curr = dict(line.split("=", 1) for line in self.remote._session.run("env")[1].splitlines())
+ self._curr = dict(line.split("=", 1) for line in self.remote._session.run("env -0")[1].split('\x00'))
```

Although I'm not sure how cross-platform the `-0` flag is for `env`...

campbellr:
Awesome, thanks @tomerfiliba (sorry about the broken 'fix' i suggested)!

pigay:
Thanks @tomerfiliba ;)

PR #179, opened 2015-02-08
Fall back to `printenv` per variable if no `env -0` (#178).

abarnert:
You may not want this fix—read #178 and decide whether the imperfect but faster solution suggested by pigay in #162 is a better fallback for non-GNU systems than this one. But you definitely want **something**, because as it stands you can't access the remote environment at all on OS X, FreeBSD, etc.

- Since `env -0` is a GNU extension, we need some fallback for OS X, etc.

- Parsing `env` into `name=value` lines (skipping the ones without `=`),
  then calling `printenv $name` on each name (skipping the ones that
  don't exist) should work portably, although obviously more slowly.

### DIFF
--- a/plumbum/machines/remote.py
+++ b/plumbum/machines/remote.py
@@ -15,8 +15,24 @@ class RemoteEnv(BaseEnv):
     __slots__ = ["_orig", "remote"]
     def __init__(self, remote):
         self.remote = remote
-        self._curr = dict(line.split("=", 1) for line in
-                self.remote._session.run("env -0; echo")[1].split("\x00") if "=" in line)
+        session = remote._session
+        # GNU env has a -0 argument; use it if present. Otherwise,
+        # fall back to calling printenv on each (possible) variable
+        # from plain env.
+        env0 = session.run("env -0; echo")
+        if env0[0] == 0 and not env0[2].rstrip():
+            self._curr = dict(line.split('=', 1)
+                              for line in env0[1].split('\x00')
+                              if '=' in line)
+        else:
+            lines = session.run("env; echo")[1].splitlines()
+            split = (line.split('=', 1) for line in lines)
+            keys = (line[0] for line in split if len(line)>1)
+            runs = ((key, session.run('printenv "%s"; echo' % key))
+                    for key in keys)
+            self._curr = dict((key, run[1].rstrip('\n')) for (key, run) in runs
+                              if run[0] == 0 and run[1].rstrip('\n')
+                              and not run[2])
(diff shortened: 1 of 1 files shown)


tomerfiliba:
do you suppose it's safe to assume python on the remote host? is so,

`python -c 'import pickle, os; print(pickle.dumps(dict(os.environ)))'`

i'd love to hear your opinion.

-tomer

-----------------------------------------------------------------

*Tomer Filiba*
tomerfiliba.com     <http://www.facebook.com/tomerfiliba>
<http://il.linkedin.com/in/tomerfiliba>

abarnert:
That's definitely safe for my currently envisioned uses, but that's mainly wanting to get your rpyc working so I can play with it. :) I don't know how other people use Plumbum. For example, I can imagine wanting to use it to ssh into my jailbroken iPhone, which has a reasonably complete BSD userland but doesn't have Python (although there is a mostly-working Python 2.5 package I could install).

But if it's acceptable, it seems like a great solution. And it also solves #180, and anything similar that comes up; Python provides a lot of functionality that's consistent across platforms that a POSIX userland doesn't guarantee. (It seems like you could even, with a bit more work, make `RemoteMachine` just work on Windows as long as they have an ssh server and a `python` on the `%PATH%`.)

One little thing: I think you want `dumps(dict(…), 2)` to force a protocol. That way a remote machine whose `python` is 3.x (as on at least Arch Linux) will still send back something you can parse on 2.3-2.7.

abarnert:
After looking around a bit more, I think you do want to use `python` for `env`, and maybe `stat`, and maybe a few other things.

First, you're already calling Python in least one place, to call `platform.uname`, so I think you've already set a precedent to rely on it (although you do try `uname` first there).

Just for fun, I tested my jailbroken iPhone, and… it has GNU coreutils, even though it's Darwin. So, switching on `uname` doesn't help.

Launching Python just to `stat` a file sounds like a terrible performance problem, but the worst I could find on any machine I have access to (my iPhone) is 64ms vs. 9ms; I think it might be acceptable.

`_path_glob` calls `echo $fn` without quotes on each filename, so any run of whitespace gets turned into a single space, and various special characters get mangled. If you fix that with `echo "$fn"`, then it mostly works—except that filenames with newlines now show up as multiple separate files. You could fix that with `printf "$fn\0"` (which is portable, unlike anything fancy with `echo`).

`_path_listdir` calls ls -a` and assumes one filename per line, so it also has problems with filenames that contain newlines. Of course you could replace this with a glob (`for f in .* *`), so if you solve the previous problem, you solve this one.

PR #197, opened 2015-07-09
Fix mismatch between format string and format tuple in remote._path_link

tillberg:
This PR just fixes a little mismatch in BaseRemoteMachine._path_link between the arguments in the format string and the format tuple. I changed the `"-s"`in the format string to a `%s` that should get replaced by `"-s" if symlink else ""` at run time, which appears to have been the original intent.


tomerfiliba:
thanks

PR #202, opened 2015-07-11
Two more modifiers for return codes

henryiii:
I was writing a little tutorial for plumbum and converting one of my more complex bash scripts to plumbum, and I noticed one common use that `subprocess.call` does well that plumbum works a little harder for. It was in checking and working with return codes. While the exception method is excellent, if you are working with, for example, git, and you want to toggle behavior based on the return code of a command (like `git rev-list HEAD..origin`), then you have to use `.run(retcode=None)[0]` or try/except blocks (which then require accessing the stdin/stderr differently in the except block).

I added two modifiers alongside FG and BG, namely TF and RETCODE, that make working with return codes easy and natural in plumbum. They work a lot like BG, for example:
```python
>>> git['rev-list', 'HEAD..origin'] & TF
True
>>> git['rev-list', 'HEAD..origin'] & TF(1)
False
>>> git['rev-list', 'HEAD..origin'] & RETCODE
0
```

They support foreground execution with `FG=True` in the call, and TF can take retcode just like FG. Note that the bash syntax for using true/false return codes is &&, so this is a little like it, like BG.

I've added the two new objects, a test, documentation, and docstrings. Also fixed one minor typo in the docstring for FG.

PS: Should I add to the "what's new" documentation, too? I'm assuming you do that, but I'm happy to help if you need me to. Also, let me know if you have recommendations about names, etc, I'm happy to change them and resubmit if you have better ideas.) Thanks for such a brilliant library! I'll send a link to my blog post tutorial when I'm done with it if you want to see it. It should have a plumbum vs. bash example.

### DIFF
--- a/docs/local_commands.rst
+++ b/docs/local_commands.rst
@@ -129,6 +129,25 @@ one you passed::
    If you wish to accept several valid exit codes, ``retcode`` may be a tuple or a list. 
    For instance, ``grep("foo", "myfile.txt", retcode = (0, 2))``   
 
+
+If you need the value of the exit code, there are two ways to do it. You can call ``.run(retcode=None)``
+(or any other valid retcode value) on a command, you will get a tuple ``(retcode, stdin, stdout)``. If
+you just need the recode, or want to check the retcode, there are two special objects that can be applied
+to your command to run it and get or test the retcode. For example::
+
+    >>> cat["non/existing.file"] & RETCODE
+    1
+    >>> cat["non/existing.file"] & TF
+    False
+    >>> cat["non/existing.file"] & TF(1)
+    True
+
+.. note::
+   If you want to run these commands in the foreground (see below), you can give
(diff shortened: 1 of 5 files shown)


tomerfiliba:
adding a what's new would be great, thanks

tomerfiliba:
btw, i'm looking for a maintainer to make more regular releases, would you be interested?

henryiii:
Sure, I'd be happy to help, just let me know what you want me to do. I'll see if I can update the what's new docs in the next couple of days.

henryiii:
My blog post is [here](http://iscinumpy.blogspot.com/2015/07/plumbum-scripting.html).

tomerfiliba:
nice post! btw, i made you a collaborator of the project (you should have push access). it would be great if you could update the docs (where applicable) and let me know when everything's right so i'd make a release.

PR #206, opened 2015-07-22
Plumbum.color

### LINKED ISSUE #128: Add terminal colors (posix/windows) to cli.terminal
see #127

henryiii:
This is the `plumbum.colors` module I've been working on, designed to make adding colors easy and safe. I've prepared a post about it [here](http://iscinumpy.blogspot.com/2015/07/plumbum-color.html),  this [post](http://iscinumpy.blogspot.com/2015/07/plumbum-scripting.html) has been updated, and there is extensive documentation in the addition. As a quick example:

```python
from plumbum import colors
print("This is red" << colors.red + "This is not red")
with (colors.blue+colors.bold):
    print("Styles can be combined, and are context managers.")
```

You can access the ANSI 16 basic colors, 256 colors (with names), and full 24 bit color - and can convert between them (This is the only library I know of that can do that). You can easily subclass Style to create more representations (HTMLStyle is included to make documentation easier to write). Styles are smart and know how to combine to create the most effective ANSI sequence. Styles correctly handle being turned off, and don't turn on for non-posix tty systems (can be forced, though). Basic colors can be used on Windows with a tool (like colorama) to convert ANSI codes. (See examples).

Docs (with list of color names), API docs, tests, examples included.

Closes #128


piotr-dobrogost:
Would it make sense to release it as a standalone package and use from within plumbum?

tomerfiliba:
Looks nice, but why is ``COLOR``? it could be just a normal module ``plumbum.color`` and then people would use it as ``from plumbum.color import red``. Also, I find the ``*`` very weird. How about brackets? E.g., instead of

    print("This is red"*COLOR.RED + "This is not red")

Use
    print(red["This is red"] + "This is not red")

And then colors could be combined with ``&``, e.g., ``(red & bold)["red and bold"]``.

This commit it huge, i'd start by putting it in ``plumbum.experimental`` and then it'd be easier to look at it.

henryiii:
If it was a stand-alone package, it couldn't a.) be integrated with plumbum.cli, and b.) it would run in to the same problem as colorama and all other color packages: they have to be installed to be used, and they are generally "non-essential", and would require a dependency that you would have to check for. If a script depends on plumbum, now it has guaranteed easy access to color.

The bracket syntax works just fine, try it. :) `(red + bold)["red and bold"]`

``*`` or ``<<`` is easier when actually writing code IMO, because you don't have to wrap it.

Instead of having an experimental directory, how about making an experimental branch, and then PRing parts into the main branch when ready? That way testing it would be easy, just checkout a branch. Putting things in a separate directory breaks all test scripts, docs, code, etc. I'll set it up if you want me to. I think we really should have two branches at least, a master that mostly takes bug fixes, and a experimental/dev that takes new features and matures them.

(PS: It's also easy to checkout my fork branch into a local branch with a different name, that's how I am working now)

henryiii:
PS: With the exception of the CLI integration, tests, and docs, everything new is contained in `plumbum.color`. If I rename the COLOR object to color, I'll need a new directory name. `style`?

henryiii:
The 16 default colors, attributes, and ``reset`` are all available in ``plumbum.color`` now. For the extended colors, rgb, hex colors, color iteration, etc. still require the ``COLOR`` object (which is still called ``COLOR`` for now).

henryiii:
I've renamed all colors and attributes to lowercase, and I've even made the module syntax work. So from `plumbum.colors import red` or even from `plumbum.colors.bg import blue` works now! (similar but slightly improved version of plumbum.cmd trick). All the non-module like features, like `with colors:` are still supported.

henryiii:
I've added access to colors to the command line, too, using
```bash
$ python -m plumbum.colors
```

This was always available as an emergency terminal recovery (no arguments is still full reset), but selecting all available styles and using colors instead of colorlib is new. tom Have you had time to look at it yet? If you need prebuilt access to the docs, they are at http://henryiiiplumbum.readthedocs.org/en/latest/.

henryiii:
Closing in favor of new PR from color branch.

PR #207, opened 2015-07-31
local.cwd doesn't work if cwd changes by non-plumbum means

henryiii:
This patch makes the `local.cwd` the current working directory, not just the current working directory when Plumbum loaded, so code like this now works:

```python
assert local.cwd == os.getcwd()
os.chdir('../plumbum')
assert local.cwd == os.getcwd()
os.chdir('../tests')
assert local.cwd == os.getcwd()
```

The `cwd` is now a descriptor in the local machine (and works from class or object access). Added a test similar to the lines above to verify that it works.

(Example of use case: This is a common problem when working with Plumbum in IPython and trying to use `local.cwd` to find files after changing the directory with `cd`.)

### DIFF
--- a/plumbum/lib.py
+++ b/plumbum/lib.py
@@ -59,3 +59,13 @@ class six(object):
         def get_method_function(m):
             return m.im_func
 
+class StaticProperty(object):
+    """This acts like a static property, allowing access via class or object.
+    This is a non-data descriptor."""
+    def __init__(self, function):
+        self._function = function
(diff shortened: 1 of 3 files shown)


tomerfiliba:
Looks good. I'd just call ``_FunctionDescriptor`` a ``StaticProperty``.

henryiii:
Done. Moved it to lib.py, too.

PR #208, opened 2015-08-20
Syntax highlighting in README

adamchainz:
Basically tell RST compiler that it's Python/bash. Github specifically makes no assumptions hence no highlighting without these.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,9 @@ Now let's see some code!
 
 Cheat Sheet
 -----------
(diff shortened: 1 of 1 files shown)


henryiii:
Looks good, I'm going to go ahead and merge, since this is a docs issue.

henryiii:
By the way, thanks for the patch!

adamchainz:
:rainbow: :smile_cat:

PR #216, opened 2015-08-30
SwitchAttr support of values via environment variables

tigrawap:


plumbum/cli/application.py
```diff
@@ -320,8 +317,39 @@ def _parse_args(self, argv):
                 else:
                     swfuncs[swinfo.func] = SwitchParseInfo(swname, (val,), index)
 
+        # Extracting arguments from environment variables
+        envindex = 0
+        for env, swinfo in self._switches_by_envar.items():
+            envindex-=1
```
tomerfiliba: what is the negative index used for?

plumbum/cli/application.py
```diff
@@ -320,8 +317,39 @@ def _parse_args(self, argv):
...
+            envval = os.environ.get(env)
+            if envval is None:
+                continue
+
+            if swinfo.func in swfuncs:
+                continue # skip if overridden by command line arguments
+
+            val = self._handle_argument(envval, swinfo, env)
+            envname = "$%s" % env
```
tomerfiliba: is env a string? if so, please use `envname = "$%s" % (env,)`

plumbum/cli/application.py
```diff
@@ -320,8 +317,39 @@ def _parse_args(self, argv):
...
+            if swinfo.list:
+                swfuncs[swinfo.func] = SwitchParseInfo(envname, ([val],), envindex)
```
tomerfiliba, line 337: it would be useful to add a comment saying you're handling only one value for lists, and perhaps add an assert to catch wrong usage (if applicable)

plumbum/cli/application.py
```diff
@@ -320,8 +317,39 @@ def _parse_args(self, argv):
                 else:
                     swfuncs[swinfo.func] = SwitchParseInfo(swname, (val,), index)
 
+        # Extracting arguments from environment variables
+        envindex = 0
+        for env, swinfo in self._switches_by_envar.items():
+            envindex-=1
+            envval = os.environ.get(env)
```
tomerfiliba: since it's plumbum (after all), you should use `local.env` instead of `os.environ`

tomerfiliba:
added some comments -- thanks!

henryiii:
Very handy idea! I can help add docs when it's ready.

plumbum/cli/application.py
tigrawap: SwitchParseInfo requires index, which later is used for sorting, I think that using same index as before will be wrong, command line arguments and environment variables arguments are not really same indexes. But since it's used later for sorting(and probably some other things that I did not notice) -  indexes are required, so negative is like "pre-command line " indexes
But maybe you have another idea how to handle this

plumbum/cli/application.py
tigrawap: Yeah string, ok, will update  
Btw, dollar here just to give proper representation, that this is not regular name, but environment variable, whenever this will be printed

plumbum/cli/application.py
tigrawap, line 337: Oh, sure, had this on commit-comment level only

plumbum/cli/application.py
henryiii: You could use `'${0}'.format(env)`, that's my preference, but plumbum already uses the % formatting in most places so that's okay.

tigrawap:
Updated based on comments, squashed, ready to merge

PR #220, opened 2015-09-08
Issue#219 - Fix for invalid stat arg on FreeBSD

astraw38:


henryiii:
What about the almost matching lines directly above it, for `_path_getuid`, `_path_getgid`?

henryiii:
Looks good to me. I'll leave it to @tomerfiliba for the final check.

PS, do you have direct access to a FreeBSD system? I'm curious as to how the default terminal works with `plumbum.colors`.

tomerfiliba:
thanks

astraw38:
@henryiii 

Only SSH access unfortunately.

henryiii:
@astraw38 Okay, thanks for checking. Thanks for the patch, too!

PR #223, opened 2015-09-17
Making path more compatible with pathlib

henryiii:
The stdlib (in newer Python 3's) has a `pathlib` module, with object oriented paths. I've been working at making Plumbum's Path more like `pathlib.Path`.

Added:
```
    .stem           # The last component of a path, without extension
    .root           # The root of the path ('/' or '\\')
    .drive          # The drive letter ('' on unix)
    .parts          # The split path, including the root
    .iterdir()      # Directory iterator, slightly faster than list on Python 3.5
    .as_uri()       # The URI of a path
```

Renamed:
```
    .isdir()  -> .is_dir()
    .islink() -> .is_symlink()
    .isfile() -> .is_file()
    .basename -> .name
```

For the renames, old names still work, but emit a `DepreciationWarning`. Also, I just assumed a remote path should have a `ftp` URI, but am open to suggestions there. Drive and root are always `''` and `'/'` for remote paths. If pathlib exists, several features, including the ones above, are tested against it.

I also rooted out and killed a bug for `skipIf` that was incorrectly masking tests on Python 2.6.


tomerfiliba:
interesting. but why do you think the URI should be ``ftp`` and not ``ssh``? or ``sftp``?

henryiii:
I'm open to that, I wasn't sure what would be most useful. I chose ftp because it is easier to add an s than remove one. Maybe `ssh` would be better for a default? I can add a parameter to set that, something like `as_uri(service='ssh')`.

PR #226, opened 2015-09-25
Better test support

henryiii:
I have added OSX builds to Travis and added Windows builds through Appveyor. Getting all builds to pass required a few changes:

- MacOS compiled without issue due to previous changes. I'm using brew to prepare the vm. I did move the build command to a script to make it easier to toggle for Mac.
- Windows required a bunch of skip tests for now. Hopefully some of them (like ones that depend on `ls`) will get fixed eventually. I did not add remote tests either (yet)
- Test skipping is a little more organized now.
- To keep Appveyor from hanging because `nosetests` is interpreted as a batch file, I moved from `nosetests` to `py.test`. It can run all `unittest` and `nosetests` out of the box, and can greatly improve future test writing in a simpler format if we want it to (and we can stop patching `unittest` in Py2.6!). For now, all tests were left alone.
- Several windows Py3 bugs for iterators and strings were fixed.


henryiii:
@tomerfiliba, you might want to set up an official appveyor account for the project, I have one monitoring my fork for testing the addition, but it's obviously HenryIII/plumbum.

tomerfiliba:
hey @henryiii , i merged in your other pull requests and now this one has a conflict

henryiii:
Fixed. There's an odd bug in Travis's Python 3.5 and dev, causing it to run the previous python when you run python,  rather than python3; causing the tunnel test to fail on all versions (including ones that passed before) (and, I believe that's what's happening, I could be wrong). I'll see if I can fix that test later.

henryiii:
I found the problem, it was a change in behavior for string arguments for Python's that were installed automatically by Travis (but not preinstalled ones). Fixed in master.

Issue #232, opened 2015-10-21
CLI subcommand "help" instructions are annoying


markerikson:
Running a cli.Application command with a "-h" currently prints out a "see 'myScript.py subcommand --help' for more info" message for every subcommand.  I personally find this to be annoying and unnecessary.  Would it be possible to make this optional somehow?

henryiii:
I can look at this in a few days, I'd be open to it. If you have any ideas on how best to do it, feel free to try it out and let us know.

markerikson:
Looks good so far.  Thanks for the quick response!

PR #233, opened 2015-10-22
Fix for extra help message

henryiii:
This small patch allows customization or removal of the extra help message on subcommands, addresses #232. You can remove a subcommand extra help message by setting the class variable `SUBCOMMAND_HELPMSG` to `False`/`None`, or you can customize it with a string (`{parent}` and `{sub}` refer to the parent and subcommand name, respectively, in the string).

Also fixed color output on subcommands, the help message color matches the subcommand color now, if the subcommand color was set. See the geet.py example for an example of both changes.

PS: Unrelated to this patch: Travis fails occasionally, simple rerunning of the failed matrix element without changes fixes the problem.

### DIFF
--- a/examples/geet.py
+++ b/examples/geet.py
@@ -49,6 +49,7 @@ except ImportError:
 from plumbum import cli, colors
 
 class Geet(cli.Application):
+    SUBCOMMAND_HELPMSG = False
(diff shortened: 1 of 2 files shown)


henryiii:
Since this fixes #232 and includes a fix to colors, too, I'll go ahead and merge.

PR #237, opened 2015-11-03
Cli direct run

henryiii:
This adds calling an `Application()` directly without arguments as a shortcut to `Application.run()`. This is needed for some systems with entry points (like conda) that do not support a `.` in the entry point syntax.


henryiii:
Note: this might conflict with my previous request, since they both change the same file, I'll fix that once you give the go ahead.

PR #239, opened 2015-11-03
Utilopen

henryiii:
This is a minor utility to allow opening files in the native viewer. This is similar to `webbroswer.open`, but according to the official docs:
> Note that on some platforms, trying to open a filename using this function, may work and start the operating system’s associated program. However, this is neither supported nor portable

This is a portable way to do it without abusing that module. It is in `plumbum.path.utils` and is called `gui_open`.

Works on Mac, Linux, Windows.


AndydeCleyre:
Oh, I like this.

Issue #240, opened 2015-11-09
plumbum hides stderr of pipes


hashbrowncipher:
```
Type "help", "copyright", "credits" or "license" for more information.
>>> from plumbum.cmd import cat, head
>>> from plumbum import FG
>>> cat['/dev/urndom'] & FG(1)
/bin/cat: /dev/urndom: No such file or directory
>>> (cat['/dev/urndom'] | head['-c', '10']) & FG(retcode=1)
>>>
```

I would expect to see the standard error output of cat printed to the stderr of the Python process.  For instance, zsh shows this:

```
❯❯❯ /bin/cat /dev/urndom | head -c10
/bin/cat: /dev/urndom: No such file or directory
```

henryiii:
I'll try to look at these soon! Hoping it's an easy fix. Feel free to dig around a bit, would start in plumbum/commands.

henryiii:
I've got a fix for this, will commit in a day or two.

hashbrowncipher:
Hey, I don't see code for this in master.  Did something prevent you from merging the change you had?

henryiii:
Seems to work...

```
maya8:plumbum henryiii$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
maya8:plumbum henryiii$ python
Python 3.5.0 |Anaconda 2.4.0 (x86_64)| (default, Oct 20 2015, 14:39:26) 
[GCC 4.2.1 (Apple Inc. build 5577)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from plumbum import FG
>>> from plumbum.cmd import cat, head
>>> cat['/dev/urndom'] & FG(1)
cat: /dev/urndom: No such file or directory
>>> (cat['/dev/urndom'] | head['-c', '10']) & FG(retcode=1)
cat: /dev/urndom: No such file or directory
```

henryiii:
Should also be tested for, now, with py.test, using file descriptor capture:
```python
    @skip_on_windows
    def test_pipe_stderr(self, capfd):
        from plumbum.cmd import cat, head
        cat['/dev/urndom'] & FG(1)
        assert 'urndom' in capfd.readouterr()[1]

        assert '' == capfd.readouterr()[1]

        (cat['/dev/urndom'] | head['-c', '10']) & FG(retcode=1)
        assert 'urndom' in capfd.readouterr()[1]
```

thomasshanks:
@henryiii:
I'm getting this new error:
[...]

PR #243, opened 2015-11-11
Ensure fair error attribution of pipelined return codes

### LINKED ISSUE #241: ProcessExecutionError is misleading when piped commands fail
Relates to #145 

```
Type "help", "copyright", "credits" or "license" for more information.
>>> from plumbum.cmd import cat, head
>>> from plumbum import FG
>>> (cat['/dev/urndom'] | head['-c', '10']) & FG()
  File "/nail/home/josnyder/consul-backups/env/local/lib/python2.7/site-packages/plumbum/commands/modifiers.py", line 143, in __rand__
    cmd(retcode = self.retcode, stdin = None, stdout = None, stderr = None)
  File "/nail/home/josnyder/consul-backups/env/local/lib/python2.7/site-packages/plumbum/commands/base.py", line 89, in __call__
    return self.run(args, **kwargs)[1]
  File "/nail/home/josnyder/consul-backups/env/local/lib/python2.7/site-packages/plumbum/commands/base.py", line 219, in run
    return p.run()
  File "/nail/home/josnyder/consul-backups/env/local/lib/python2.7/site-packages/plumbum/commands/base.py", line 181, in runner
    return run_proc(p, retcode, timeout)
  File "/nail/home/josnyder/consul-backups/env/local/lib/python2.7/site-packages/plumbum/commands/processes.py", line 217, in run_proc
    return _check_process(proc, retcode, timeout, stdout, stderr)
  File "/nail/home/josnyder/consul-backups/env/local/lib/python2.7/site-packages/plumbum/commands/processes.py", line 34, in _check_process
    stdout, stderr)
plumbum.commands.processes.ProcessExecutionError: Command line: ['/usr/bin/head', '-c', '10']
Exit code: 1
```

In this example, the process which failed is the cat process.  The exception message is misleading, because it implies that the head process failed.

hashbrowncipher:
Fixes #241 by making the (new) verify method on command objects responsible for determining whether to raise an exception.

### DIFF
--- a/plumbum/commands/base.py
+++ b/plumbum/commands/base.py
@@ -5,6 +5,7 @@ from plumbum.commands.processes import run_proc, iter_lines
 from plumbum.lib import six
 from tempfile import TemporaryFile
 from subprocess import PIPE, Popen
+from types import MethodType
 
 
 class RedirectionError(Exception):
@@ -296,6 +297,16 @@ class Pipeline(BaseCommand):
             dstproc.returncode = rc_src or rc_dst
             return dstproc.returncode
         dstproc.wait = wait2
+
+        dstproc_verify = dstproc.verify
+        def verify(proc, retcode, timeout, stdout, stderr):
+            #TODO: right now it's impossible to specify different expected
+            # return codes for different stages of the pipeline, but we
+            # should make that possible.
+            proc.srcproc.verify(retcode, timeout, stdout, stderr)
+            dstproc_verify(retcode, timeout, stdout, stderr)
+        dstproc.verify = MethodType(verify, dstproc)
+
         return dstproc
 
 class BaseRedirection(BaseCommand):
--- a/plumbum/commands/processes.py
+++ b/plumbum/commands/processes.py
@@ -20,18 +20,7 @@ except ImportError:
 # utility functions
 #===================================================================================================
 def _check_process(proc, retcode, timeout, stdout, stderr):
-    if getattr(proc, "_timed_out", False):
-        raise ProcessTimedOut("Process did not terminate within %s seconds" % (timeout,),
-            getattr(proc, "argv", None))
-
-    if retcode is not None:
-        if hasattr(retcode, "__contains__"):
-            if proc.returncode not in retcode:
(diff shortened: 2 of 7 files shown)


hashbrowncipher:
Worth noting is that my penultimate commit modifies a test that would otherwise fail due to a behavioral change.  I believe that fixing the test, as it was originally written, would require implementing the TODO I left on line 303 of base.py.

henryiii:
This looks good to me. I'll let @tomerfiliba or @koreno check it over, then I'm okay to merge. I will need this merged before I submit my fix for #240.

henryiii:
I'll go ahead and merge this, so I can add a few help strings and submit my patch.

henryiii:
We'll need to fix that test before releasing 1.6.1, I believe, as only requiring the final error code seems more intuitive (and it's the way it used to work). I'm looking at that, feel free to do so too.

henryiii:
I believe the issue really is that plumbum follows a non-standard return code policy. That is,
`false | true` has a return code of 0 in bash. Plumbum uses `or` to combine return code, making 0 special regardless of your requested return code.

I've set it so that 0 is expected for everything except the last command, which fixes the test that failed because it is the way plumbum has always worked. It just didn't attribute the non-zero command because it was a simple or, rather than the method of this patch.

If we can attach an expected return code to processes, that would be the ideal way to allow custom return code management. That should not be hard now that each command is properly tested.

henryiii:
Correction: I've now set it so that the old behavior is exactly restored (though with the correct attribution now in place). We can discuss changing it for 1.7.0, but probably not for v1.6.1. I think adding a way to set each command's return code is probably the way to go from here, though. Like:
```python
ls['-a'].set_retcode([0,1]) & FG
```
Where `.set_retcode` sets a property and returns `self`. Then, if that property exists / is not some special value, that is used for verify instead. A little verbose, but don't see another better way ATM.

hashbrowncipher:
That's gorgeous, thanks @henryiii.

I think your solution involving set_retcode is a good one.  I would recommend normative names like 'expect_code' or 'should_return' or (if you like brevity) 'returns'.

PR #244, opened 2015-11-13
Py.test conversion

henryiii:
This patch completely converts the error system to native pytest. This has several changes in syntax that I think make test writing cleaner and simpler, hopefully encouraging the writing of new tests. It also completely removed the hacks we used to have that make unittest and nose behave on Python 2.6. I've spent too much time debugging those to want to have to continue to have them. We've been using pytest since 1.6.0, but have been using the old syntax on the old tests (new tests were in the new syntax).

* Tests use a simple assert statement instead of a method
  - Missing Py 2.6 assert methods are no longer needed
  - Descriptive error reporting for all comparisons, including `in` (missing in Py 2.6)
* Tests use `with pytest.raises` to handle exceptions
  - No need for lambda functions or odd syntax, use normal calls
  - All raises tests are much cleaner
* Built in skip function (missing in unittest for Py 2.6)
  - No longer needed to be patched in by plumbum!
* Proper capture of stdout
  - Can even capture file descriptors, allowing access to stderr/stdout of other programs run by python
  - No need to manually specify capture, it is assumed by the use of the proper fixtures
* Easy to skip tests based on importability
* Clean skip/warning/error reporting
* No need to subclass a special class, pytest uses discovery
* Simple functions can be added too, if they don't fit a category

This patch also includes the very simple fix to #240, the test for that patch is the reason I finally converted the error system.


henryiii:
This fix also fixes the windows build; includes a small change to `.parts` on windows including the drive to match pathlib.

henryiii:
The expanded tests are already discovering bugs, and it's harder to maintain two active branches, so I'll merge this if no one has an objection. Since I added coverage on both branches, it's clear that the coverage is the same (that is, the new syntax didn't lose any tests in conversion).

PR #246, opened 2015-11-23
Check the returncode of sshpass when reporting errors

vodik:
sshpass does error reporting with its returncode:

 1 Invalid command line argument
 2 Conflicting arguments given
 3 General runtime error
 4 Unrecognized response from ssh (parse error)
 5 Invalid/incorrect password
 6 Host public key is unknown. sshpass exits without confirming the new key.

This patch tries to check the return code on error cases and reports an
appropriate error, so in the case of an invalid login or missing host
public key, we get a useful error message instead of a generic
SSHCommsError.

### DIFF
--- a/plumbum/machines/session.py
+++ b/plumbum/machines/session.py
@@ -19,6 +19,12 @@ class SSHCommsError(EOFError):
 class SSHCommsChannel2Error(SSHCommsError):
     """Raises when channel 2 (stderr) is not available"""
 
+class IncorrectLogin(SSHCommsError):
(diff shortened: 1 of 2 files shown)


vodik:
Now added some tests as well, but they're not so useful for the continuous integration since if you can't install sshpass inside of it.

henryiii:
Looks pretty good to me.

vodik:
Just tweaked `IncorrectLogin` and `HostPublicKeyUnknown` to subclass `SSHCommsError`

PR #268, opened 2016-03-25
Simplify implementation of TEE modifier

AndrewPashkin:
Subj.

### DIFF
--- a/plumbum/commands/modifiers.py
+++ b/plumbum/commands/modifiers.py
@@ -1,6 +1,6 @@
 import os
 from select import select
-from subprocess import PIPE
+from subprocess import STDOUT
 import sys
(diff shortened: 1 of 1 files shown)


henryiii:
I still haven't checked this to make sure it works...

henryiii:
This does not seem to actually print to stdout and save, like TEE is supposed to. Reopen if you can show what I'm doing wrong! :)

I just tried:
```python
In [1]: from plumbum import local, TEE
In [2]: echo = local['echo']
In [3]: echo['This is fun'] & TEE
This is fun
Out[3]: (0, 'This is fun\n', '')
```

The `This is fun` line about `Out[3]` is missing with this patch.

PR #273, opened 2016-04-10
use % formatter

EdwardBetts:


henryiii:
Thanks, I'll see if I can merge this as soon as I fix the Travis builds. Unfortunately, there's an unstable test (the nohup one) that causes the builds to sporadically fail (I think).

plumbum/cli/progress.py
```diff
@@ -139,16 +139,16 @@ def __str__(self):
         if width - len(ending) < 10 or self.has_output:
             self.width = 0
             if self.timer:
-                return "{0:g}% complete: {1}".format(100*percent, self.str_time_remaining())
+                return "{0:%} complete: {1}".format(percent, self.str_time_remaining())
```
henryiii: This is slightly different, this returns 0.000000% instead of 0% (etc), it should be: `{0:.0%}` to be more reasonable.

plumbum/cli/progress.py
```diff
@@ -139,16 +139,16 @@ def __str__(self):
...
             else:
-                return "{0:g}% complete".format(100*percent)
+                return "{0:%} complete".format(percent)
```
henryiii: Same.

henryiii:
Yep, the % formatter is in the Py 2.6 docs, so it can be merged. I had one comment, though, can you fix that first?

EdwardBetts:
I made the extra change.

PR #281, opened 2016-06-09
Add timeouts into the ExecutionModifiers

fahhem:
This pipes run_proc's timeout argument farther up the stack.

I'm hoping the tests will catch if this hits any parts that don't accept timeout and I'll add it to those as well.

### DIFF
--- a/plumbum/commands/modifiers.py
+++ b/plumbum/commands/modifiers.py
@@ -2,6 +2,7 @@ import os
 from select import select
 from subprocess import PIPE
 import sys
+import time
 from itertools import chain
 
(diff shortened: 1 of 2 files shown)


fahhem:
Looks like the only failure in travis is due to Python nightly changing the string representation of ints. Should I fix that in this PR or a separate one?

henryiii:
One PR is okay with me.
Henry

fahhem:
Okay, the latest push should have passing tests

henryiii:
Okay, I’ll look at it when I get off work.
Thanks!
Henry

PR #288, opened 2016-07-25
Apply retcode=None to the whole Pipeline tree (ie. when >2 commands in pipeline)

xpinguin:
Hi!

I've noticed, that current implementation does not allow to disable returncode verification for the whole pipeline (Pipeline tree),
eg. `(netstat["-tnlp"] | grep[pid] | grep[port])(retcode = None)` fails at grep[pid]: nonzero returncode
while `(netstat["-tnlp"] | grep[pid])(retcode = None)` works as expected

Am I missing something?

Thanks! :)


henryiii:
Thanks! I'll merge it and then write some tests.

PR #308, opened 2017-01-31
Scrap redundant encoding

koreno:

PR #309, opened 2017-02-01
Revert "Scrap redundant encoding"

henryiii:
Reverts tomerfiliba/plumbum#308

I added the `__contains__` method to `base.py` a few months ago, and the encoding changes break all travis tests.

Leaving the new encoding test.


henryiii:
@koreno Let me know if you have a way to simplify the encoding part, the original method seems to be the only way to be compatible with both Py2 and Py3.

henryiii:
Nevermind, you were correct, @koreno. It was just the new test that fails, in both cases. Fixed on master to be an expected fail, one of us should check into why it fails...

koreno:
Yeah I need to take another look at this...

PR #313, opened 2017-02-27
[FIX] fix missing parent in __init__ method for subclass when using --help-all

sebastienbeau:
Hi,
We have a project"voodoo" that use plumbum and the option --help-all do not work
The issue is that the "--help-all" method add the parent after calling the __init__ method of my subclass.

Here my code that need parent in __init__ https://github.com/akretion/voodoo/blob/master/voodoo/main.py#L149

The changed proposed fix the issue

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -47,6 +47,12 @@ class GeetAdd(cli.Application):
 class GeetCommit(cli.Application):
     message = cli.Flag("-m", str)
 
+    def __init__(self, *args, **kwargs):
+        # This code is here to test that the help-all
+        # is passing the parent in the init
+        super(GeetCommit, self).__init__(*args, **kwargs)
+        self.is_parent_debug = self.parent.debug
+
     def main(self):
         if self.parent.debug:
             return "committing in debug"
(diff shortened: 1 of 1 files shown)


henryiii:
Could you provide a little example of what this fixes? It looks quite reasonable, but not sure how to break it. I tried to run the geet.py example and it seems to be fine with and without this patch. I'd like to make a test for it, this is what I tried:

```python
def test_helpall(self, capsys):
     _, rc = Geet.run(["geet.py", "--help-all"], exit=False)
     assert rc == 0
     stdout, stderr = capsys.readouterr()

     assert "-m" in stdout
     assert "--version" in stdout
     assert "add" in stdout
     assert "commit" in stdout
```

But it works with and without the patch.

plumbum/cli/application.py
```diff
@@ -578,8 +578,9 @@ def helpall(self):
 
         if self._subcommands:
             for name, subcls in sorted(self._subcommands.items()):
-                subapp = (subcls.get())("%s %s" % (self.PROGNAME, name))
-                subapp.parent = self
+                subapps = subcls.get()
+                subapps.parent = self
+                subapp = subapps("%s %s" % (self.PROGNAME, name))
```
henryiii: Feel free to update this to the new format syntax, too, would be cleaner:

```python
sub app = subapps("{0} {1}".format(self.PROGNAME, name))
```

(Either is a big readability improvement on the old line, though 👍 )

sebastienbeau:
Hum, it's seem that it work... Really strange, I will take the time to look at it.
On my side I am not impacted anymore (I have remove the need of calling the parent in __init__ for other reason). But at that version https://github.com/akretion/voodoo/blob/2.4.0/voodoo/main.py#L145 it was broken.
For now I close the PR and I will re-open it with a test that fail

PR #317, opened 2017-04-07
Make argument type error messages use `__str__` method of `Predicate`s

gmarceau:
Previous behavior looks like this
```
Error: Argument of output expected to be <plumbum.cli.switches.Predicate object at 0x10ce91c10>, not 'requirements.txt':
        ValueError('output is a file, should be a directory',)
```
now looks like this
```
    Error: Argument of output expected to be 'MakeDirectory', not 'requirements.txt':
        ValueError('output is a file, should be a directory',)
```
when checking a `Predicate`.

### DIFF
--- a/plumbum/cli/application.py
+++ b/plumbum/cli/application.py
@@ -349,8 +349,8 @@ class Application(object):
                 return argtype(val)
             except (TypeError, ValueError):
                 ex = sys.exc_info()[1]  # compat
-                raise WrongArgumentType("Argument of %s expected to be %r, not %r:\n    %r" % (
-                    name, str(argtype), val, ex))
+                raise WrongArgumentType("Argument of {name} expected to be {argtype}, not {val!r}:\n    {ex!r}".format(
+                    name=name, argtype=argtype, val=val, ex=ex))
         else:
             return NotImplemented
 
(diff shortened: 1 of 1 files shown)


henryiii:
Would you mind changing this to use `"{0} {1}".format()` style or `"{val} {other}".format()` style? I'd prefer that going forward.

plumbum/cli/application.py
```diff
@@ -350,7 +350,7 @@ def _handle_argument(val, argtype, name):
             except (TypeError, ValueError):
                 ex = sys.exc_info()[1]  # compat
                 raise WrongArgumentType("Argument of %s expected to be %r, not %r:\n    %r" % (
-                    name, argtype, val, ex))
+                    name, str(argtype), val, ex))
```
henryiii: Something like:
```python
raise WrongArgumentType("Argument of {name} expected to be {argtype}, not {val!r}:\n    {ex!r}".format(
                    name=name, argtype=argtype, val=val, ex=ex))
```

gmarceau: That works

henryiii:
As soon as Travis finishes, I'll accept.

gmarceau:
We've been using a ton of plumbum at bodylabs. It is replacing both our usages of argparse and of Bash, no less. Thank you for creating it.

PR #318, opened 2017-04-07
Add `MakeDirectory` Predicate

gmarceau:


plumbum/cli/switches.py
```diff
@@ -1,5 +1,6 @@
 from plumbum.lib import six, getdoc
 from plumbum import local
+from plumbum.cmd import mkdir
```
henryiii: path's already have a mkdir.

plumbum/cli/switches.py
```diff
@@ -416,6 +417,15 @@ def ExistingDirectory(val):
         raise ValueError("%r is not a directory" % (val,))
     return p
 
+
+@Predicate
+def MakeDirectory(val):
+    p = local.path(val)
+    if p.is_file():
+        raise ValueError('%r is a file, should be nonexistent, or a directory' % (val,))
+    mkdir('-p', p)
```
henryiii: `p.mkdir()` should work here.

henryiii: That is embarrassing; I was going to ask you to add a test for this, and was going to show you an example, but apparently none of the predicates are tested. 😕 So I'll take this without tests and I'll add them for all four later (unless you'd like to!)

PR #325, opened 2017-05-25
Pathlib like

henryiii:
Minor additions to add path lib like syntax. Added paths to quickref.


piotr-dobrogost:
Couldn't pathlib just be vendored in?

henryiii:
That would be great, but there are several things that make that difficult:

* Plumbum has remote paths. That doesn't exist in Pathlib
* In Plumbum, all paths are absolute
* Plumbum has a lot of custom methods that would need to be ported in (at least for a while, some would stay)
* Plumbum has to support Py2 for a while, which would mean that the back ported pathlib2 should be part of Plumbum (the 2.6 support that was "recently" added helps, licenses is OK, I believe)

The first bullet point is the hardest, and the others would just take work. I don't have time, but would be okay to help someone interested.

piotr-dobrogost:
*The first bullet point is the hardest (…)*

Do you have any idea how to go about integrating this functionality with pathlib?

henryiii:
I have a rough plan if interested:
1. Plumbum could provide a subclass of Path that would add the remote functionality.
2. All paths generated by Plumbum would simply be absolute (resolved). That would be pretty easy.
3. In the same way as 1, Plumbum would provide a subclass that would add the extra methods Plumbum provides for local paths.

A few other points:
* Plumbum parts should remain as separated as possible (modular). So the parts that use paths should only use official methods wherever possible. If the more useful methods (or even remote paths) could eventually be added to pathlib, that would allow Plumbum to drop them completely.
* This could initially be called `plumbum.pathlib`
* The differences sheet in the QuickRef should be filled out to make transition easier.
* If you got, say, `plumbum.cli` without the rest of plumbum, it should be able to use pathlib

Of course, I mean pathlib2 on older versions of Python.

henryiii:
This could be extended to add remote paths:

inheritance diagram

Pathlib is pretty simple, and is [a single file](https://github.com/mcmtroffaes/pathlib2/blob/develop/pathlib2.py). The remote paths should be general enough to not specifically depend on Plumbum (possibly in a non-concrete and concrete version?)

PR #326, opened 2017-06-05
Adding support for GSS-API (Paramiko 1.15 and above needed)

capean:
Added for Kerberos authentication. Please let me know if anything is incorrect, since this is my first pull request.


henryiii:
LGTM. If the CI build works, I'm happy to merge. Normally I like to see a test added as well, but for something like kerberos auth, I don't think a test is needed.

henryiii:
Thanks!

capean:
Thank you

PR #328, opened 2017-06-15
Fix #327

### LINKED ISSUE #327: "type" parameter to prompt defaults to int, but docs say str
[terminal.py:111](https://github.com/tomerfiliba/plumbum/blob/df7ab6a543a0f7046a01240259669ed5b971a17c/plumbum/cli/terminal.py#L111):

```python
def prompt(question, type = int, default = NotImplemented, validator = lambda val: True):
```

[terminal.py:117](https://github.com/tomerfiliba/plumbum/blob/df7ab6a543a0f7046a01240259669ed5b971a17c/plumbum/cli/terminal.py#L117):

```
    :param type: The type of the answer, defaults to str
```

One of these should change, though I'm not sure which one. My guess would be make the default `str`.

AndydeCleyre:
Fix #327 in one of two obvious ways. prompt type defaults to str, to match docs.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,7 @@
 -------------------
 
 * Fixes for globbing with spaces in filename on a remote server (`#322 <https://github.com/tomerfiliba/plumbum/issues/322>`_)
+* Terminal: Changed ``prompt()``'s default value for ``type`` parameter from ``int`` to ``str`` to match existing docs (`#327 <https://github.com/tomerfiliba/plumbum/issues/327>`_)
 
 1.6.3
 -----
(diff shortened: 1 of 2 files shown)


henryiii:
Can you add a note to the changelog, too, since this could possibly change code for someone who depends on the default parameter? Otherwise I think this is the correct way to change it.

AndydeCleyre:
Let me know if it's a problem that my editor murdered a bunch of trailing spaces in the changelog.

henryiii:
Trailing spaces deserve to be murdered :)

henryiii:
Thanks!

AndydeCleyre:
Thank you, this continues to be an amazing project.

PR #339, opened 2017-08-14
Add localization functions and russian localization

### LINKED ISSUE #335: I18N of cli.Application
It would be nice to have `--help` and usage messages localized, maybe using GNU gettext?

ASMfreaK:
This PR fixes #335 - It adds gettext localization to all strings that are used in Application output.
Also as a test I've written russian localization. Any other language can be easily added. 
This should be tested on py2.x and on Windows.


henryiii:
This looks great. I'll try to get it in in a day or two. Thanks!

henryiii:
Would you mind changing the `%` formatting into `{name}` formatting? That would be more consistent than using both styles, and would be easier for translators to use (for example, if they need to reorder a pair of messages). The fact that Plumbum currently mixes both styles is a historical relic.

ASMfreaK:
I like braces formatting, too. Ok.

henryiii:
LGTM.

henryiii:
Thanks!!!

ASMfreaK:
Yay~

PR #347, opened 2017-09-21
Fix #293: AttributeError: 'SshMachine' object has no attribute 'env'

### LINKED ISSUE #293: AttributeError: 'SshMachine' object has no attribute 'env'
I'm trying to Ssh on a remote machine, and get following error:

```
  File "wakeonlan", line 49, in <module>
    main()
  File "wakeonlan", line 44, in main
    wake_remotely(args.host)
  File "wakeonlan", line 30, in wake_remotely
    globals()[func_name]()
  File "wakeonlan", line 22, in remote_wake_huckfinn
    with SshMachine("host", user="user") as rem:
  File "/usr/lib/python3.5/site-packages/plumbum/machines/ssh_machine.py", line 106, in __init__
    new_session = new_session)
  File "/usr/lib/python3.5/site-packages/plumbum/machines/remote.py", line 164, in __init__
    self.env = RemoteEnv(self)
  File "/usr/lib/python3.5/site-packages/plumbum/machines/remote.py", line 38, in __init__
    BaseEnv.__init__(self, self.remote.path, ":")
  File "/usr/lib/python3.5/site-packages/plumbum/machines/env.py", line 37, in __init__
    self._update_path()
  File "/usr/lib/python3.5/site-packages/plumbum/machines/env.py", line 40, in _update_path
    self._path.update(self.get("PATH", ""))
  File "/usr/lib/python3.5/site-packages/plumbum/machines/env.py", line 24, in update
    self[:] = [self._path_factory(p) for p in text.split(self._pathsep)]
  File "/usr/lib/python3.5/site-packages/plumbum/machines/env.py", line 24, in <listcomp>
    self[:] = [self._path_factory(p) for p in text.split(self._pathsep)]
  File "/usr/lib/python3.5/site-packages/plumbum/machines/remote.py", line 202, in path
    p = self.env.expanduser(p)
AttributeError: 'SshMachine' object has no attribute 'env'
```

The error occurs on creation of the SshMachine:

```
with SshMachine("host", user="user") as rem:
    pass
```

It can be easily reproduced when you put some path with ~ into the PATH environment variable on the remote host:

`export PATH=$PATH":~/bin:"`

I used pyhon 3.5.2 with plumbum 1.6.2

rrei:
As described in my comment, I just moved the code to `BaseRemoteMachine` and called those methods from `RemoteEnv` :)

### DIFF
--- a/plumbum/lib.py
+++ b/plumbum/lib.py
@@ -6,12 +6,11 @@ import inspect
 
 IS_WIN32 = (sys.platform == "win32")
 
-def _setdoc(super, name=None):  # @ReservedAssignment
+def _setdoc(super):  # @ReservedAssignment
     """This inherits the docs on the current class. Not really needed for Python 3.5,
     due to new behavoir of inspect.getdoc, but still doesn't hurt."""
     def deco(func):
-        func_name = name or func.__name__
-        func.__doc__ = getattr(getattr(super, func_name, None), "__doc__", None)
+        func.__doc__ = getattr(getattr(super, func.__name__, None), "__doc__", None)
         return func
     return deco
 
--- a/plumbum/lib.py
+++ b/plumbum/lib.py
@@ -6,11 +6,12 @@ import inspect
 
 IS_WIN32 = (sys.platform == "win32")
 
-def _setdoc(super):  # @ReservedAssignment
+def _setdoc(super, name=None):  # @ReservedAssignment
     """This inherits the docs on the current class. Not really needed for Python 3.5,
     due to new behavoir of inspect.getdoc, but still doesn't hurt."""
     def deco(func):
-        func.__doc__ = getattr(getattr(super, func.__name__, None), "__doc__", None)
+        func_name = name or func.__name__
+        func.__doc__ = getattr(getattr(super, func_name, None), "__doc__", None)
         return func
     return deco
 
--- a/plumbum/machines/remote.py
+++ b/plumbum/machines/remote.py
@@ -55,26 +55,13 @@ class RemoteEnv(BaseEnv):
         self.remote._session.run("export " +
             " ".join("%s=%s" % (k, shquote(v)) for k, v in self.getdict().items()))
 
+    @_setdoc(BaseRemoteMachine, name="_expand")
     def expand(self, expr):
-        """Expands any environment variables and home shortcuts found in ``expr``
-        (like ``os.path.expanduser`` combined with ``os.path.expandvars``)
+        return self.remote._expand(expr)
 
-        :param expr: An expression containing environment variables (as ``$FOO``) or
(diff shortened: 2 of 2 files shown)


henryiii:
No, I think is due to the `__doc__` not being writable in Python2. I think it's an easy fix, let me know if you need help.

I'm also going to need to fix the Windows test, that's been failing for a long time due to an re-ordering issue that's not related to this patch.

henryiii:
PS: There might be a Plumbum decorator that does what you want.

henryiii:
Try something like this:
```python
@_setdoc(BaseRemoteMachine)
def expand(self, expr):
    self.remote.expand(expr)
```

And remove the `_` in front of the methods for the remotes. (if you *really* think the `_` are good ideas, we can modify `_setdoc` to take an optional name, it's in `plumbum/lib.py`.

henryiii:
PS: Also take a look at the tests in `tests/test_remote.py`, it would be great if you could add a test or two. Otherwise I'll have to and that could delay things. ;)

rrei:
Alright, thanks. I'll fix this and take a look at the tests.

rrei:
I'll have to move the RemoteEnv class till after BaseRemoteMachine for the decorator to work though. Hope that doesn't break anything.

henryiii:
Try it and see...

henryiii:
Or you could reverse the one that has the doc string and the one with the decorator...

rrei:
Yes, that makes much more sense than moving a whole class :D
I'm just struggling to see how I could test this. I'd have to prepare the target machine with a PATH entry having "~"... any suggestions?

henryiii:
We might not be able to change the path easily before connecting, but wouldn't this also be needed after connecting? So the path could be changed after connecting, then used?

henryiii:
Tried it locally, doesn't seem to fail even on the old version. So might not be trivial to test with travis. Unless you have an idea, I'm okay with it. Please check locally and make sure it fixes your use case, at least.

rrei:
It's not happening in your case because the remote PATH doesn't contain any entries with '~'.
I've checked and this solves the issue for me, and probably the same was happening with `SshMachine`.

rrei:
Still clueless on how to force the remote to have a "~" PATH entry in an automated test... :D

henryiii:
LGTM, thanks!

rrei:
Hi! Is there any date planned for the next release? I'd like to include plumbum straight from PyPI in my dependencies :)

henryiii:
I'd like to get a few more translations  in, have a few people I might be able to ask. Probably 2 weeks?

rrei:
Alright, great news! Thanks for the update and for building such a great library ;)

henryiii:
It was a bit late, but 1.6.4 is now officially out!

PR #351, opened 2017-10-14
First go at Dutch translation.

roelaaij:
The trickiest to translate was is the word for "switch", using the Dutch translation of "switch as in light switch", which is "knop", doesn't sound right, as that is generally used as a translation for "button" . "Schakelaar" would be another possibility, but that refers to a switch object more than to the concept of a switch. In the end I choose "optie", which is also the translation of "option".


henryiii:
Looks great! I'm taking this opportunity to drop 3.3 support as well, since it seems like it's not supported by the optional packages used in Travis tests. That should get the build to finish.

Since argument (`--this=that`) was translated to a different word than flag (`--this`), I think "optie" is clear.

henryiii:
Thanks!

PR #355, opened 2017-11-16
Make it possible to write to stdin of a pipeline.

hashbrowncipher:
This enables the use-case where a Python process takes a file object as
argument. For instance (with boto3):

    with cmd.bgrun(stdin=PIPE) as f:
        boto3.download_fileobj(bucket, key, f.proc.stdin)


hashbrowncipher:
I think this is worthwhile, but the code is not ready. I'll PR again when I have something that passes tests.

henryiii:
Okay. Will review it then.

PR #356, opened 2017-11-17
Make it possible to write to stdin of a pipeline (try #2)

hashbrowncipher:
This enables the use-case where a Python function takes a file object as
argument. For instance (with boto3):

    with cmd.bgrun(stdin=PIPE) as f:
        boto3.download_fileobj(bucket, key, f.proc.stdin)

The first PR (#355) broke functionality because it changed the behavior of Popen objects. Specifically, the dstproc Popen object had a non-None stdin field that it didn't understand how to deal with. This change avoids that problem by not exposing a stdin to the underlying Popen object. The only way I could think to do this was by switching from inheritance (in IterablePopen) to composition.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -12,7 +12,8 @@ requirements:
 
   run:
     - python
-    - paramiko
+    - paramiko<2.4 # [py26]
+    - paramiko # [not py26]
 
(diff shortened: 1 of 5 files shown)


henryiii:
Thanks! LGTM.

PR #360, opened 2017-12-05
Added .editorconfig

KOLANICH:


henryiii:
Python2 has changed on macOS, I'll fix that soon, don't worry about the failing tests. Might have to drop Py 2.6 also, will try not to though.

KOLANICH:
@henryiii, Here is th actual build matrix https://github.com/KOLANICH/RichConsole/blob/master/.travis.yml

henryiii:
Thanks! The build has been fixed, as well now.

PR #361, opened 2017-12-05
Added main module detection and cli construction, now should show the…

KOLANICH:
… right name (instead of __main__.py) when called for modules


henryiii:
Needs Python 2 support and a test, do you think you could add those? (I might be able to do the Py2 support if you add a test.) If a test is not possible, then at least a little example would help.

KOLANICH:
Sorry, I'm not going to add them, I have no time for it. But I have tested that against my apps using plumbum.cli, they work as indended. If we strictly need tests,  something like https://github.com/laffra/auger (an automatic tests generator) may be helpful.

KOLANICH:
About python2: how 'bout dropping support of  it?

henryiii:
For a end-user library, sure; but for something like Plumbum it can't be done, at least for a few more years. Plumbum needs to work anywhere a shell script works, and by default, CentOS7 and macOS still only provide Python 2. And I also work for LHCb, and they are not supporting Python 3 for another year or two. Rewriting 1M+ lines of code is not easy. I can't even drop Python 2.6 support yet since it is default on SL6 😢

henryiii:
I'll look at adding tests then, but it will be a while before I have time.

plumbum/cli/application.py
```diff
@@ -166,7 +168,11 @@ def __init__(self, executable):
         # Filter colors
 
         if self.PROGNAME is None:
-            self.PROGNAME = os.path.basename(executable)
+            spec=get_main_module_frame().f_globals['__spec__']
```
henryiii:
```suggestion
            spec=get_main_module_frame().f_globals.get('__spec__', None)
```

henryiii:
Wouldn't it have been easier to address the Py2 + test request than to maintain this as a branch for 3+ years? Anyway, Python 2 support is working with one small change.

Currently the only problem is that this breaks the current tests on Windows - it seems it is replacing the program name with `python -m __main__`?

KOLANICH:
I have looked at the code once again, it looks like it is incorrect.

1. it looks like it would break if not `__main__` filename is used
2. about `python` - to properly get a name of the executable I had to use `ctypes`:

```python
import ctypes
import platform
impl=platform.python_implementation()
if impl=="cpython" or impl == "PyPy" :
  ctypes.pythonapi.Py_GetArgcArgv.restype = None
  ctypes.pythonapi.Py_GetArgcArgv.argtypes = [
    ctypes.POINTER(ctypes.c_int),
    ctypes.POINTER(ctypes.POINTER(ctypes.c_wchar_p))
  ]
  def realArgv():
    count = ctypes.c_int()
    args = ctypes.pointer(ctypes.c_wchar_p())
    ctypes.pythonapi.Py_GetArgcArgv(ctypes.byref(count), ctypes.byref(args))
    argc = count.value
    argv = [str(args[i]) for i in range(count.value)]
    return argv
print(realArgv())
```

but it is not a solution that can be used, since it would break for the impls without ctypes

henryiii:
Could you add a test? I think it's safe to merge now that Py 2 is gone.

henryiii:
Also you can use the original simpler solution, I think, now without Py 2. Is there still an issue on Windows?

henryiii:
```
assert 'geet add' in "Usage:\n    python -m __main__
```

Seems to be doing the opposite on Windows?

KOLANICH:
I usually don't use Windows, and I don't know when I use it. 
We surely need automated tests for that, and GH CI can be used to test on different platforms (though the code I have written is cross-platform). I'm sorry. To be honest I'm currently occupied by other things, I just saw your merge commit (it seems you cannot force-push into my repo) in notifications and so just rebased my commit over master and force-pushed, since it doesn't require much refocusing.

henryiii:
Okay. I tried to get this working briefly before release, then saw the regression on Windows and left it. I don't use Windows much either, I'm just going from the GHA CI (we do have automated tests here). We also need a test showing the "working" behavior so we can verify it works there too. I'll eventually try to come back by and see if I can fix it, then.

PR #366, opened 2017-12-22
image.py: fix old-style(?) relative import statement

### LINKED ISSUE #365: SyntaxError in plumbum/cli/image.py
image.py contains the line

import . as cli

All versions of python I tried (3.6, 3.5, 2.7) raise SyntaxError reading this line. According to the Python Language Reference, such syntax of the import statement is not allowed. What exactly is the purpose of this line? And how to achieve this purpose in legal python?

tkluck:
This fixes #365. I can't actually find when this syntax became illegal, but this replacement works both in Python 2.7 and 3.6 whereas the original doesn't.

As I wrote in #365, this patch unbreaks installs under versions of `pip` that have a hard dependency on the `py->pyc` step succeeding.

### DIFF
--- a/plumbum/cli/image.py
+++ b/plumbum/cli/image.py
@@ -2,7 +2,7 @@ from __future__ import print_function, division
 
 from plumbum import colors
 from .termsize import get_terminal_size
-import . as cli
+from .. import cli
 import sys
 
 class Image(object):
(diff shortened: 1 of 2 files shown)


henryiii:
Just to make sure that this doesn't break again, I'll add an import for image to the tests.

henryiii:
THAnks!

tkluck:
Thanks @henryiii ! Can I request a (point) release for this one? It's blocking my install in production.

henryiii:
Sure. I’m traveling for Christmas so might not be immediate. I’d like to fix the slow load/added requirement from the translation addition too. ~1 week?

tkluck:
That would be marvellous. Thanks! Let me know if I can help.

henryiii:
Done, github and pypi are one 1.6.5, and conda-forge should follow soon.

PR #368, opened 2018-01-06
Handle remotes that do not us a compatible default shell

DragonSA:


plumbum/machines/ssh_machine.py
```diff
@@ -113,16 +119,31 @@ def popen(self, args, ssh_opts = (), **kwargs):
...
         cmdline.extend(ssh_opts)
         cmdline.append(self._fqhost)
-        if args and hasattr(self, "env"):
-            envdelta = self.env.getdelta()
-            cmdline.extend(["cd", str(self.cwd), "&&"])
-            if envdelta:
-                cmdline.append("env")
-                cmdline.extend("%s=%s" % (k, v) for k, v in envdelta.items())
+        remote_cmdline = []
+        if shell_cmd:
```
henryiii, line 123: This is undefined and throws an error when run. What is `shell_cmd`?

plumbum/machines/ssh_machine.py
```diff
@@ -113,16 +119,31 @@ def popen(self, args, ssh_opts = (), **kwargs):
...
+            formulate = arg.formulate() if isinstance(arg, BaseCommand) else (shquote(arg), )
```
henryiii, line 124: `arg` is also undefined. Only `args` exists at this point.

henryiii (commented):
Looks like I never submitted this review, but GH saved it as pending!

Issue #369, opened 2018-01-13
PuttyMachine upload/download fails with non-default port


kelbyers:
When creating a PuttyMachine with a port parameter passed, the value of `port` is not passed on to the `SshMachine` initializer to be used for the scp command, and no scp_opts are configured.

When SshMachine sets up arguments for scp_machine, port is None and scp_opts are none, so `pscp` gets called without an argument to use the port that `pssh` is using, and attempts to upload or download files using the ParamiknoMachine fail.

https://github.com/tomerfiliba/plumbum/blob/4ea9e4c00a8d36900071393954d2df90931ad689/plumbum/machines/ssh_machine.py#L277-L282

```
In [9]: mach = PuttyMachine('diskstation', port=922)

In [10]: (mach['date'])()
Out[10]: 'Sat Jan 13 11:55:03 CST 2018\n'

In [11]: (mach['ls'])()
Out[11]: 'modules.conf\n'
In [12]: mach.download('modules.conf','.')
---------------------------------------------------------------------------
ProcessExecutionError                     Traceback (most recent call last)
<ipython-input-12-8f0bbf2c7ef2> in <module>()
----> 1 mach.download('modules.conf','.')

c:\users\kel\appdata\local\temp\pb\lib\site-packages\plumbum\machines\ssh_machine.py in download(self, src, dst)
    241             src = self._translate_drive_letter(src)
    242             dst = self._translate_drive_letter(dst)
--> 243         self._scp_command("%s:%s" % (self._fqhost, shquote(src)), dst)
    244
    245     @_setdoc(BaseRemoteMachine)

c:\users\kel\appdata\local\temp\pb\lib\site-packages\plumbum\commands\base.py in __call__(self, *args, **kwargs)
     93     def __call__(self, *args, **kwargs):
     94         """A shortcut for `run(args)`, returning only the process' stdout"""
---> 95         return self.run(args, **kwargs)[1]
     96
     97     def _get_encoding(self):

c:\users\kel\appdata\local\temp\pb\lib\site-packages\plumbum\commands\base.py in run(self, args, **kwargs)
    223         """
    224         with self.bgrun(args, **kwargs) as p:
--> 225             return p.run()
    226
    227

c:\users\kel\appdata\local\temp\pb\lib\site-packages\plumbum\commands\base.py in runner()
    185             was_run[0] = True
    186             try:
--> 187                 return run_proc(p, retcode, timeout)
    188             finally:
[...]

PR #370, opened 2018-01-13
Support PuttyMachine file upload and download using non-default port

kelbyers:
This is a fix for issue #369. If a PuttyMachine instance is created to an ssh server that is running on the non-default port, then the upload and download methods fail, because `pscp` is still using the default port instead of the actual port.

I have added some tests. The tests verify:
- PuttyMachine calls the `__init__` of SshMachine with the correct parameters
- PuttyMachine includes the correct port option for pscp with non-default ports
- `PuttyMachine.__str__` has the expected message.

The tests can run on any system even without Putty installed. One of the test fails on PyPy, but it is marked with `@xfail_on_pypy`. It fails on PyPy because the feature of `pytest-mock` used in that test does not work on PyPy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,13 +14,13 @@ matrix:
       env: PYV=3.6
 #    - python: nightly
     - python: pypy
-    - language: generic
-      env: PY3=2 PYV=Mac2
-      os: osx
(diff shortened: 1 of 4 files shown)


henryiii:
Looks great! Thanks

PR #378, opened 2018-03-13
Introduce DESCRIPTION_MORE attribute

acehreli:
This change adds a new attribute to be printed between DESCRIPTION and USAGE.

Paragraphs are determined by empty lines inside DESCRIPTION_MORE. Paragraph indentations are preserved and lines are wrapped. Additionally, extra spaces are removed from text.


acehreli:
Closing this one as I'm not happy with how paragraphs and wrapping are handled.

acehreli:
We've been using this change in production for about three weeks now.

plumbum/cli/application.py
```diff
@@ -618,6 +625,108 @@ def help(self):  # @ReservedAssignment
...
+                        if not paragraph:
+                            # Start a new paragraph
+                            paragraph = line
+                            initial_indent = indent
+                            subsequent_indent = indent
+                        else:
+                            # Add to current paragraph
+                            paragraph = paragraph + ' ' + line
+
+            yield from current()
```
henryiii: Your production must not use Python 2 (must be nice...). This `yield from` should be very easy to convert to:

```python
for item in current()
    yield item
```

Since it is an iterator and not a generator.

plumbum/cli/application.py
```diff
@@ -618,6 +625,108 @@ def help(self):  # @ReservedAssignment
...
+                """
+                if paragraph:
+                    yield paragraph, initial_indent, subsequent_indent
+
+            for part in text.lstrip("\n").split("\n"):
+                indent, line = split_indentation(part)
+
+                if len(line) == 0:
+                    # Starting a new paragraph
+                    yield from current()
```
henryiii: Same here as below.

acehreli:
Thanks, henryiii...

PR #379, opened 2018-03-13
English consistency

acehreli:
Fixes mostly minor issues with the English used in /cli.


henryiii:
(fixing the Travis builds due to recent change in brew, should be able to merge after that)

henryiii:
By the way, the one change I'm not very fond of is subcommand -> sub-command; since subcommand makes a good method/class name, and sub-command is not valid python or C++.

acehreli:
Thanks for the feedback. I don't have a strong opinion between sub-command and subcommand but after Googling for both, sub-command was a clear winner. :) And subcommand is not in the dictionaries that I checked.

If we prefer subcommand, I propose changing existing sub-commands to subcommand because my main point was consistency.

henryiii:
I think consistency is more important; we'll go with sub-command. Still waiting on the Travis fixes, brew's default Python3 is confusing Travis. Hope to have it working soon.

henryiii:
Thanks!!!

PR #386, opened 2018-04-09
Add run_MODIFIER methods to the command object

TMiguelT:
e.g., allows you to `ls["-a"].run_bg()` instead of `ls["-a"] & BG`. This is a more pythonic/explicit syntax for this common use-case


TMiguelT:
Thanks for merging!

PR #391, opened 2018-05-21
reformat code with yapf

FooBarQuaxx:
The project is pretty awesome. The only think that's missing is good pep8ish compliant code formatting.


henryiii:
This sounds like a great idea! What's the command to run it? Might make sense as a CI check as well.

FooBarQuaxx:
The tools is called yapf: https://github.com/google/yapf

FooBarQuaxx:
Thanks for accept my contribution.

PR #401, opened 2018-08-07
allow switch abbrevations as an option

presto8:
One feature I missed from argparse is the ability to have abbreviations or partial switches (such as "--foo" matching "--foobar", assuming no ambiguous conflicts). This patch attempts to add that. I also added a unit test and it seems to pass. I'm not sure if the authors are in favor of this feature, but sharing it in case it would be beneficial to others.

P.S. I looked around in the code and docs for an already existing implementation, I hope I didn't miss it and do redundant work :-)

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -424,6 +424,27 @@ specify the validators. For example::
 
 Annotations are ignored if the positional decorator is present.
     
+Switch Abbreviations
+^^^^^^^^^^^^^^^^^^^^
+
+The cli supports switches which have been abbreviated by the user, for example, "--h", "--he", or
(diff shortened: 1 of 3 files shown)


henryiii:
Nice, as long as it's optional (it is) I like it. Would you like to add a note to the docs?

presto8:
Thanks! Sure, let me update the docs, rebase the change and update the pull request.

presto8:
I added some docs, I tried to match the style of the existing docs. And that is a good point about typos!

henryiii:
Looks good to me, thanks!

presto8:
Thank you, as well, for all of your hard work on this great project!

PR #406, opened 2018-09-11
Implement RemotePath.open for ParamikoMachine-associated paths

smheidrich:
While ``LocalPath.open`` [is implemented](https://github.com/tomerfiliba/plumbum/blob/22db2e82b114c02592716530241448bb66b76fdf/plumbum/path/local.py#L223-L225) and works fine, ``RemotePath.open`` is currently an [empty placeholder method](https://github.com/tomerfiliba/plumbum/blob/22db2e82b114c02592716530241448bb66b76fdf/plumbum/path/remote.py#L296-L297). However, for ``ParamikoMachine``-associated paths, an implementation is trivial due to the existence of [``SFTPClient.file``](http://docs.paramiko.org/en/latest/api/sftp.html#paramiko.sftp_client.SFTPClient.file), which does pretty much exactly what we want.

This PR exploits that fact and offers a simple implementation of ``RemotePath.open`` which uses ``SFTPClient.file`` for ``ParamikoMachine``-associated paths<sup>1</sup> and raises ``NotImplementedError`` otherwise (a change from the previous behaviour of just returning ``None``).

<sup>1</sup> technically also anything else which has a ``sftp.file`` attribute on its ``remote``

### DIFF
--- a/plumbum/path/remote.py
+++ b/plumbum/path/remote.py
@@ -293,8 +293,18 @@ class RemotePath(Path):
                 "got %r" % (dst, ))
         self.remote._path_link(self, dst, True)
 
-    def open(self):
-        pass
+    def open(self, mode="r", bufsize=-1):
+        """
+        Opens this path as a file.
+
+        Only works for ParamikoMachine-associated paths for now.
+        """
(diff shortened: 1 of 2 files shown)


henryiii:
Improvement to ParamikoMachine always welcome!

PR #410, opened 2018-09-25
Towards pathlib.Path.mkdir compatibility

smheidrich:
This PR aims to make the ``mkdir`` method of Plumbum's path classes (local & remote) more compatible with [``pathlib.Path.mkdir``](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir) by implementing the arguments ``parents`` and ``exist_ok`` that are present in the latter. The ``mode`` argument has also been added to the argument list to match the signature of pathlib's method and preserve backwards-compatibility once it is implemented in the future, but I haven't implemented it yet, because pathlib does some umask-voodoo that I'd have to look into in more detail. I might add it later.

Should be backwards-compatible with earlier Plumbum ``mkdir`` behavior (which unfortunately required choosing defaults different from pathlib's, however) and of course comes with tests.

### DIFF
--- a/plumbum/machines/remote.py
+++ b/plumbum/machines/remote.py
@@ -380,8 +380,10 @@ class BaseRemoteMachine(BaseMachine):
     def _path_copy(self, src, dst):
         self._session.run("cp -r %s %s" % (shquote(src), shquote(dst)))
 
-    def _path_mkdir(self, fn):
-        self._session.run("mkdir -p %s" % (shquote(fn), ))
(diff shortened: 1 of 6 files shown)


henryiii:
I'd like pathlib compatibility (I'd love it if someday remotes were just subclassed pathlib paths), but we probably shouldn't make a backward incompatible default change until 1.7.0, for example.

smheidrich:
Thanks for merging! :+1:

PR #418, opened 2018-10-15
Fix file handle leak in NOHUP / posix_daemonize

algrs:
posix_daemonize failed to close a file handle, which lead to resource exhaustion if you NOHUP'd many thousands of subprocesses.

This commit adds a test demonstrating the leak and closes the appropriate file.


henryiii:
LGTM, thanks for adding a test!

PR #420, opened 2018-10-23
Implement mode argument for LocalPath.mkdir

smheidrich:
As a continuation of #410, this PR implements the "mode" argument for ``LocalPath.mkdir``, which turned out to be trivial, as the umask stuff that is mentioned in ``pathlib.Path.mkdir``'s docs also applies to ordinary ``os.mkdir`` (it actually goes back to the ``mkdir`` syscall).

It's still not implemented for ``RemotePath.mkdir``, because the ``mkdir`` shell command (used by Plumbum's remote mkdir implementation) handles the ``--mode`` argument differently, namely without the umask transformation done by the ``mkdir`` syscall, so it's a bit more involved. We'd have to retrieve the umask of the remote process manually in order to do that transformation, but I'm not sure whether one should then also be able to set the remote umask and how that should work in detail. So I'll look into it again later.

The docstring is formulated in such a way that it leaves open the possibility of just not implementing the umask handling for remote paths (only local ones), if it turns out that it doesn't make sense.

### DIFF
--- a/plumbum/path/base.py
+++ b/plumbum/path/base.py
@@ -268,12 +268,15 @@ class Path(str, six.ABC):
         """
         Creates a directory at this path.
 
-        :param mode: **Not implemented yet,** just here to keep the order of arguments
-        backwards-compatible once it is.
+        :param mode: **Currently only implemented for local paths!** Numeric mode to use for directory
+                     creation, which may be ignored on some systems. The current implementation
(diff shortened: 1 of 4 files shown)


henryiii:
Thanks!

smheidrich:
Thanks for merging :)

PR #421, opened 2018-10-24
Docs: update remark on required parentheses

### LINKED ISSUE #412: Docs report different output than execution
https://plumbum.readthedocs.io/en/latest/local_commands.html#input-output-redirection says 

> `grep["world"] < sys.stdin > "tmp.txt"` would result in False

but in Plumbum 1.6.7 on Python 3.6.5 I get 

`TypeError: '>' not supported between instances of '_io.TextIOWrapper' and 'str'`

This seems like a good thing -- I'd much rather get an exception than a silent mistake. 

Am I misreading the docs here?

smheidrich:
Mention Python 3 behavior, explain in more detail what is going on, and fix typo.

Fixes #412.

### DIFF
--- a/docs/local_commands.rst
+++ b/docs/local_commands.rst
@@ -107,8 +107,10 @@ the output to a file named ``tmp.txt``::
     ''
 
 .. note::
-   Parenthesis are required here! ``grep["world"] < sys.stdin > "tmp.txt"`` would 
-   result in ``False``...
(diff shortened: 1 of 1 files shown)


henryiii:
Looks great! Will merge after CI builds (out of habit, since they don’t check docs really)

henryiii:
Thanks!

smheidrich:
Thanks for merging! :+1:

Issue #425, opened 2018-11-05
Stream binary data from stdout


brianbienvenu:
I'm making a Flask app which passes around large binary blobs (up to 250MB). It might transparently decompress a bz2 (I'm doing this with `bz2.BZ2File`), or it might pass that binary data through a local console application for other processing. I'm trying to use plumbum for the latter.

I can almost get what I want using `chain.popen().iter_lines()` or `chain.proc.iter_lines()`, but these are clearly intended to return a string line-by-line, instead of a binary object.

As an example, A `bz2.BZ2File()` returns a file-like object, which can be streamed out:

    >>> z = bz2.BZ2File('test.bz2')
    >>> for l in z:
    ...     print(l)
    ...
    b'\x01\x00\x01\x00"\xa4\x00\x000\x00\x00\x00\x06\x00\x00\x00ra...'
    b'\x00\x0b\x00\x0c\x00\r\x00\x0e\x00\x0f\x00\x10\x00\x11...'

This is the closest to data streaming that I've gotten out of plumbum so far:

    >>> chain = (bzcat < 'test.bz2') & BG
    >>> for l in chain.proc.iter_lines(linesize=1):
    >>>    print(l)
    ['\x01', None]
    ['\x00', None]
    ['\x01', None]
    ['\x00', None]
    ['"', None]
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/usr/local/lib/python3.7/dist-packages/plumbum/commands/processes.py", line 271, in iter_lines
        for t, line in _iter_lines(proc, decode, linesize):
      File "/usr/local/lib/python3.7/dist-packages/plumbum/commands/processes.py", line 50, in _iter_lines
        for ret in selector():
      File "/usr/local/lib/python3.7/dist-packages/plumbum/commands/processes.py", line 48, in selector
        yield key.data, decode(key.fileobj.readline(linesize))
      File "/usr/local/lib/python3.7/dist-packages/plumbum/commands/processes.py", line 264, in <lambda>
        decode = lambda s: s.decode(encoding).rstrip()
    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa4 in position 0: invalid start byte

The data processing rate is about the same as the internet download speed, so I would prefer to not call `chain()` and wait seconds for the full response to get buffered.

How can I stream data out of stdout?

smheidrich:
[...]

PR #427, opened 2018-11-09
Some more links in docs on local commands

smheidrich:
Pretty self-explanatory; I noticed that some more links to the API docs could be helpful while answering #425.


henryiii:
Looks good, thanks!

smheidrich:
Thanks for merging! :+1:

Issue #395, opened 2018-07-25
TEE loses part of process' output


ZvikaZ:
When the called command has large output, and called with TEE mode, some the of the output is lost.

See these 2 files, for example:

`long_output.py`:

     #!/usr/bin/python

    for i in xrange(4000):
        print i

And `problem.py`:

     #!/usr/bin/python

    import plumbum
    from plumbum import TEE

    exe = plumbum.local["./long_output.py"]
    ps = exe & TEE

    print "\n\n================================\n\n"

    (retcode, stdout, stderr) = ps
    with open('problem.log', 'w') as f:
        f.write(stdout + '\n')

Every time I run it, the outcome is a little bit different. But on all occasions I don't get all the 4000 lines that `long_output.py` produces.

I assume that some internal buffer holding `stdout` just get overflowed.
This could be solved by adding an option to directly write all `stdout` to some file.
It's a very common use case, and it's quite similar to Unix` `tee`.

Other possibility would be to do dynamically increase the buffer, maybe up to some (configurable) threshold - upon which some error, or at least warning, would be issued.

henryiii:
The problem is (I believe) in the direct usage of pipes - running into the warning given in the Python subprocess docs. I won't have time to look into this for a little while.

PR #434, opened 2018-12-29
Fix bug #395.

nelhage:
Even if the child process has exited, it may have written data that is
still in the pipe buffer. We need to make sure to drain it with one
final round of `select()`ing.

Note that this differs from the shell semantics of `| tee file.out`;
In a shell pipeline, the right-hand-side has no direct reference to
the left-hand-side, and so `tee` reads until it gets an `EOF` on the
pipe, normally caused by all write-side ends being closed. We can
demonstrate this with a pipeline that causes the write end to outlive
the left-hand child:

```
( sleep 500 & disown; echo hi ) | tee out.log
```

After the `hi` is printed, we can observe that the left-hand `bash` is
dead, but `tee` still has not returned:

```
$ pstree -p 7273
bash(7273)───tee(1790)
```

Killing the `sleep` makes the whole pipeline exit.

I chose to preserve the original `plumbum` semantics of "exit on child
exit" instead of the shell semantics, but wanted to call out the
difference.

### DIFF
--- a/plumbum/commands/modifiers.py
+++ b/plumbum/commands/modifiers.py
@@ -192,25 +192,43 @@ class _TEE(ExecutionModifier):
             err = p.stderr
             buffers = {out: outbuf, err: errbuf}
             tee_to = {out: sys.stdout, err: sys.stderr}
-            while p.poll() is None:
-                ready, _, _ = select((out, err), (), ())
-                for fd in ready:
-                    buf = buffers[fd]
-                    data, text = read_fd_decode_safely(fd, 4096)
-                    if not data:  # eof
-                        continue
+            done = False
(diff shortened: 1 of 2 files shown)


nelhage:
I speculate this also may explain the failure in https://github.com/tomerfiliba/plumbum/blob/eb396965a76593e2d625aaa85fcfd7f2fee03d6f/tests/test_local.py#L523-L524 but I'm not sure.

henryiii:
CI failure is not because of this PR, but because it's getting to be very hard to support CI with Python 2.6... I'll probably merge tomorrow! Looks good. Will look into removing the xfail if this fixes that error!

PR #450, opened 2019-06-26
machine.cmd

koreno:
Provide a more elegant syntax for running commands:
`machine.cmd.ls("-la")`


henryiii:
I like the symmetry with plumbum.cmd. Looks good to me!

PR #451, opened 2019-06-26
added TypedEnv, for easy parsing of env-vars

koreno:


henryiii:
Python 2 support, please. We haven't dropped Python 2.6 yet, though that's planned for 1.7. A future 2.0 could drop Python 2 support. Since Plumbum is a scripting helper, it should work with any system Python, and CentOS 7 still is on 2.6, macOS is on 2.7 (until it drops all scripting languages in a release or two, anyway).

henryiii:
I *think* you could just move the `**` test into a test marked with Python3 only.

henryiii:
Would you be interested in writing docs/changelog for this and the other additions you've just made? If not I will, but it would be slightly delayed, probably.

koreno:
> Would you be interested in writing docs/changelog for this and the other additions you've just made? If not I will, but it would be slightly delayed, probably.

I'll try to do it, unless you get to it before me...

koreno:
@henryiii - should I merge this?

henryiii:
It’s okay, I can. Have been busy so ping me as needed (like you did).

PR #454, opened 2019-07-04
Pipe to logger - a convenience mixin for sending command output to a logger

koreno:


koreno:
Another very useful pattern, for your review @henryiii

koreno:
let's put this and #455 and release 1.6.8?

PR #455, opened 2019-07-14
paths: make subscription work for composing paths

koreno:


koreno:
This is ready for you @henryiii, let me know what you think

koreno:
@henryiii - what do you say, can we merge this?
and what's the criteria for releasing a new version?

henryiii:
Sorry for the delay. I wasn't completely sure about this, I would like to make a paths behave as close to pathlib as possible,  having a many ways to do something can be confusing  and adds extra maintenance.

henryiii:
I don't think it's necessarily bad, but those were my concerns.

koreno:
> Sorry for the delay. I wasn't completely sure about this, I would like to make a paths behave as close to pathlib as possible, having a many ways to do something can be confusing and adds extra maintenance.

I guess it takes a "merge" to get your attention... :smile: 
I hear your concern, I just find the original syntax extremely inconvenient and not useful. as for pathlib - the fact it doesn't inherit from str makes it pretty inconvenient when used with other functions... but its also good since it means that subscription isn't conflicting with some other behavior (Afaik).

PR #462, opened 2019-09-09
Better documented behavoir for override, drop overwrite

### LINKED ISSUE #245: Override
While working on extending the test coverage, I noticed that the `override` parameter of `.copy` didn't actually do anything; it deleted the file before copying to the new location, but the underlying copy operation copied even if the file existed, so it was meaningless. I've fixed it to throw an error in the pytest branch, but that raises a further question: we should support a single method for choosing what is done on copy/move/etc, so should it be: 
- an override parameter, that can be `True`, `False`, or `None`, throwing an error if `False`, and silently ignoring the failed copy/move if `None`. The default would be `True`, since the current `False` acts like `True`. (Open to discussion!). This would be uniformly implemented on all copy and move functions.
- Other suggestions?

henryiii:
Fixes #245. Please comment on whether this is a good plan, or provide suggestions.

Same behavior as current Plumbum if no arguments supplied.

### DIFF
--- a/plumbum/path/base.py
+++ b/plumbum/path/base.py
@@ -260,7 +260,7 @@ class Path(str, six.ABC):
         return self.move(self.up() / newname)
 
     @abstractmethod
-    def copy(self, dst, override=False):
+    def copy(self, dst, override=None):
         """Copies this path (recursively, if a directory) to the destination path "dst".
         Raises TypeError if dst exists and override is False.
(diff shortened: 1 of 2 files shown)


AndydeCleyre:
Here there is a mismatch in `base.py`:

```python
def copy(self, dst, override=False):
```

> Will silently fail to copy if override is None (the default).

PR #464, opened 2019-09-15
Feature/proper rst

AndydeCleyre:
Small corrections for reStructuredText syntax: use a real section element rather than a horizontal line in the changelog, and replace markdown syntax in the contributing doc with that of rst.


henryiii:
Thanks!

PR #469, opened 2019-09-23
Small touchup for readme

### LINKED ISSUE #463: First two code lines on the front page
Do you think it is wise to put these two lines at the top?
```
>>> from plumbum import local
>>> ls = local["ls"]
```

These two lines are the first on the page.

I read them, and think "I don't this. This looks ugly".

Later the reader sees an easier solution.

But maybe some already left after reading the first two lines.

henryiii:
Fixes #463. Does this look like a good idea to everyone?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,8 +63,8 @@ Basics
     >>> notepad()                                   # Notepad window pops up
     u''                                             # Notepad window is closed by user, command returns
 
-In the example above, `ls = local["ls"]` is also valid and looks in your PATH; the `local` object represents your local machine.
-The `plumbum.cmd` is a useful shortcut for accessing programs in the PATH.
+In the example above, you can use `local.cmd.ls` as well, or `ls = local["ls"]` if you have an unusually named executable or a full path to an executable. The `local` object represents your local machine.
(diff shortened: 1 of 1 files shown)


koreno:
what about the new `local.cmd.ls` form? I think it's a bit more pythonic than the `from plumbum.cmd import ls`...

henryiii:
We should at least mention it, though importing things from packages is quite pythonic - I would say the `local.cmd.ls` form is handier, since you don't need multiple imports, but not more pythonic. It also downgrades more elegantly into the `local[]` form.

henryiii:
How about that? Or we can change further if needed.

koreno:
> though importing things from packages is quite pythonic

Yes, when that package actually exists... a magical `import` is much less common than a `__getattr__`.

> It also downgrades more elegantly into the local[] form.
Indeed, and also to the `RemoteMachine` case:
```
from plumbum import SshMachine
remote = SshMachine(...)
remote.cmd.ls()
```
the `import` syntax doesn't help here...

henryiii:
Okay, you've convinced me with the remote machine. I've made this a bit more obvious and reduced the magic import to a side note.

koreno:
👍😏

PR #470, opened 2019-09-26
Pre-commit support and formatting

henryiii:
Adds the PSF's Black formatter, and uses pre-commit for development. Also pyupgrade, isort, some opt-in checks to be worked on later.


henryiii:
(Will merge after 1.6.8)

PR #492, opened 2020-01-20
Path.resolve() was missing 'self'

koreno:


henryiii:
Sorry for being slow on these, I was on vacation when they came through, and forgot to get back to them.

PR #493, opened 2020-01-20
Typed env docs

koreno:
@AndydeCleyre - maybe you can comment on these docs?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -118,6 +118,7 @@ you read it in order. A quick :ref:`reference guide is available <guide-quickref
    remote
    utils
    cli
+   typed_env
    colors
    changelog
    quickref
--- a/docs/typed_env.rst
+++ b/docs/typed_env.rst
@@ -1,0 +1,96 @@
+.. _guide-typed-env:
+
+TypedEnv
+========
+Plumbum provides this utility class to facilitate working with environment variables.
+Similar to how :class:`plumbum.cli.Application` parses command line arguments into pythonic data types,
+:class:`plumbum.typed_env.TypedEnv` parses environment variables:
+
+class MyEnv(TypedEnv):
+    username = TypedEnv.Str("USER", default='anonymous')
+    path = TypedEnv.CSV("PATH", separator=":", type=local.path)
+    tmp = TypedEnv.Str(["TMP", "TEMP"])  # support 'fallback' var-names
+    is_travis = TypedEnv.Bool("TRAVIS", default=False)  # True is 'yes/true/1' (case-insensitive)
+
+We can now instantiate this class to access its attributes::
+
+    >>> env = MyEnv()
+    >>> env.username
+    'ofer'
+
+    >>> env.path
+    [<LocalPath /home/ofer/bin>,
+     <LocalPath /usr/local/bin>,
+     <LocalPath /usr/local/sbin>,
+     <LocalPath /usr/sbin>,
+     <LocalPath /usr/bin>,
+     <LocalPath /sbin>,
+     <LocalPath /bin>]
+
+    >>> env.tmp
+    Traceback (most recent call last):
+      [...]
+    KeyError: 'TMP'
+
+    >>> env.is_travis
+    False
+
+Finally, our ``TypedEnv`` object allows us ad-hoc access to the rest of the environment variables, using dot-notation::
+
+    >>> env.HOME
+    '/home/ofer'
+
+We can also update the environment via our ``TypedEnv`` object:
+
+    >>> env.tmp = "/tmp"
+    >>> env.tmp
+    '/tmp'
+
+    >>> from os import environ
+    >>> env.TMP
+    '/tmp'
+
+    >>> env.is_travis = True
+    >>> env.TRAVIS
+    'yes'
+
+    >>> env.path = [local.path("/a"), local.path("/b")]
+    >>> env.PATH
+    '/a:/b'
+
+
+TypedEnv as an Abstraction Layer
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The ``TypedEnv`` class is very useful for separating your application from the actual environment variables.
+It provides a layer where parsing and normalizing can take place in a centralized fashion.
+
+For example, you might start with this simple implementation::
+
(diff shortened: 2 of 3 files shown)


AndydeCleyre:
@koreno Thank you! That's definitely very helpful.

Some things that still confuse me about `TypedEnv`:

- I don't think the rules used for `TypedEnv.Bool` are documented anywhere. Which of these would be treated as either `True` or `False`? `yes`, `no`, `0`, `1`, `true`, `false`, `True`, `False`, `TRUE`, `FALSE`, etc.
- When using `local.env[name]`, a `KeyError` seems natural, but when using `typedenv.name`, an `AttributeError` seems to fit better, a `KeyError` seems to come out of nowhere since we're not accessing key-style, but attribute-style
- Exactly when are the env values evaluated? Those docs seem to indicate at time of access (not at time of `TypedEnv` class definition or instantiation), but the following behavior seems to contradict that:

```python
from plumbum.typed_env import TypedEnv
from plumbum import local

class AppEnv(TypedEnv):
    somenum = TypedEnv.Int("SOMENUM")
    
ae = AppEnv()
local.env['SOMENUM'] = 5
ae.somenum
```

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-6-49f2d6b17817> in <module>
----> 1 ae.somenum

/usr/lib/python3.8/site-packages/plumbum/typed_env.py in __get__(self, instance, owner)
     52                 return self
     53             try:
---> 54                 return self.convert(instance._raw_get(*self.names))
     55             except KeyError:
     56                 if self.default is NO_DEFAULT:

/usr/lib/python3.8/site-packages/plumbum/typed_env.py in _raw_get(self, *key_names)
    118                 return value
    119         else:
--> 120             raise KeyError(key_names[0])
    121 
    122     def __contains__(self, key):

KeyError: 'SOMENUM'
```

Re-instantiating `ae` also doesn't pick up on the value. Actually redefining and then re-instantiating still doesn't pick up on it, while `local.env` does see the value. Even creating a newly named class at this point and instantiating that doesn't pick up on the value. Is it frozen to the environment state at time of python process instantiation?

Also, do you think it would be a good idea for `TypedEnv` to support dict-like `.get` access for fallback values, or is the idea to just use one of (`default` during the class definition; `@property` during class definition; `try`/`except` during variable access) for that kind of behavior?

koreno:
Thanks for the comments!

> * rules used for `TypedEnv.Bool`
I will add those to the docs.

> * `AttributeError` seems to fit better
You might be right, however I did want to make it clear that there is "more than meets the eye"...
Maybe a new `class EnvironmentVariableError(AttributeError)` exception?

> * when are the env values evaluated
They indeed are evaluated at access time. It is `local.env` that's actually "freezing" the environment and copying it internally. I'm not sure why that is the implementation, it predates me...
If you want changes to be visible, use `os.environ` instead.

AndydeCleyre:
Thank you, the feature is very nice, and I appreciate the docs additions.

Yeah, as long as `isinstance(`the exception`, AttributeError)` I think that's great. And providing info or a clue that it's about env vars seems wise :+1: . I don't know if that kind of change needs to wait for a bigger version release. @henryiii -- any thoughts on the exception type or policy for changing it regarding versions?

On the evaluation time and `local.env` behavior -- oh, of course! Whoops. But it still doesn't seem right that this doesn't work:

```python
from plumbum import local
from plumbum.typed_env import TypedEnv


class AppEnv(TypedEnv):
    somenum = TypedEnv.Int("SOMENUM")

ae = AppEnv()

with local.env(SOMENUM=50):
    print(local.env["SOMENUM"])
    print(ae.somenum)
```

```
50
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-1-f3811e69ac87> in <module>
     10 with local.env(SOMENUM=50):
     11     print(local.env["SOMENUM"])
---> 12     print(ae.somenum)
     13 

/usr/lib/python3.8/site-packages/plumbum/typed_env.py in __get__(self, instance, owner)
     52                 return self
     53             try:
---> 54                 return self.convert(instance._raw_get(*self.names))
     55             except KeyError:
     56                 if self.default is NO_DEFAULT:

/usr/lib/python3.8/site-packages/plumbum/typed_env.py in _raw_get(self, *key_names)
    118                 return value
    119         else:
--> 120             raise KeyError(key_names[0])
    121 
    122     def __contains__(self, key):

KeyError: 'SOMENUM'
```

koreno:
> as long as `isinstance(`the exception`, AttributeError)` I think that's great

Unfortunately using `AttributeError` messes up the implementation (it actually indicates to python that it should try the `__getattr__`). I've made it inherit from `KeyError`, since this is still about looking up a name in a namespace, in the general sense. The exception is more to show clear error message than for use in `try/except` blocks, since normally you'd use a `default` if you expect the variable to be missing sometimes.



> support dict-like `.get` access

Indeed there is a `.get(name, default=None)` method on `TypedEnv`

AndydeCleyre:
I thought maybe instantiating `ae` within the `with local.env` block would make `SOMENUM` available to `ae`, but it didn't. Then I tried putting the whole class definition in that block, too, but it's still a `KeyError`. It should be documented if and how one can share state between `local.env` and `TypedEnv`.

koreno:
to be honest, this sharing of state between local.env and TypedEnv never came up in my work. Can you elaborate on the use-case you're looking to solve? I do agree it doesn't feel "complete", but I fear breaking things that work in an effort to fix something that won't actually be used.

My usage of TypedEnv (the reason I created it) was to centralize environment variable definition and parsing for an app, not so much as means to pass environment on to other subprocs (local.env is used when running commands via the `local` object)

AndydeCleyre:
`local.env` has been *the* way of working with environment variables in plumbum, so if using `TypedEnv` means that the user shouldn't be using `local.env` at all anymore, or that if they do they need to consider it totally incompatible with and isolated from uses of `TypedEnv`, that should be made a big loud clear warning.

koreno:
I'll clarify that in the docs, good point.

henryiii:
Is this ready for review&merge?

koreno:
> Is this ready for review&merge?

It is now

PR #495, opened 2020-02-03
Fix context manager for LocalCommand.popen()

yayachiken:
This fixes a regression introduced in 108342ad9f7580bd0ada0cf100ddf7620bf73f9e.

It appears that this commit tried to implement a transparent proxy object for `subprocess.Popen`. Unfortunately, special method calls  (e.g. of context manager methods by the `with` statement) generally circumvent the `__getattribute__` machinery and perform direct lookup.

This leads to an exception when calling LocalCommand.popen() within a with statement:

``` python
>>> from plumbum import local
>>> ls = local['ls']
>>> with ls.popen():
...     print("yay")
... 
AttributeError: __enter__
```

This broke some code of mine. I have not found any documentation regarding this change, so I assume dropping the context manager functionality was an accident.

This PR restores the context manager functionality for `local[...].popen()` calls.

### DIFF
--- a/tests/test_local.py
+++ b/tests/test_local.py
@@ -8,7 +8,7 @@ from plumbum import (local, LocalPath, FG, BG, TF, RETCODE, ERROUT, TEE,
                     CommandNotFound, ProcessExecutionError, ProcessTimedOut, ProcessLineTimedOut)
 from plumbum.lib import six, IS_WIN32
 from plumbum.fs.atomic import AtomicFile, AtomicCounterFile, PidFile
-from plumbum.machines.local import LocalCommand
+from plumbum.machines.local import LocalCommand, PlumbumLocalPopen
 from plumbum.path import RelativePath
 import plumbum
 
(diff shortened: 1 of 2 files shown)


tests/test_local.py
```diff
@@ -23,6 +23,16 @@
 SDIR = os.path.dirname(os.path.abspath(__file__))
 
 
+class TestLocalPopen:
```
henryiii, line 26: Can you stick a Python 3 guard here, please?

Also, this can just be a function, rather than a class in a function (up to you).

henryiii: Sorry I have to ask for Python 3 guards in 2020, but plumbum still supports Python 2.6. :/ But hey, we might be able to drop Python 2.6 after November 2020 (RHEL 6)!

yayachiken: It it okay like this? Or do you want me to put a run-time check on the actual methods?

(I don't know what would be appropriate behaviour then, I guess just raising `NotImplemented` instead of `AttributeError`?)

henryiii: This is fine, adding features to Python 2 is not needed, it was an error before too, I believe.

henryiii:
Thanks!

yayachiken:
Thank you, that was quick!

henryiii:
Adding tests speeds up the process considerably. ;)

PR #497, opened 2020-02-07
Moved the metadata into setup.cfg

KOLANICH:


setup.cfg
```diff
@@ -1,5 +1,56 @@
...
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Topic :: Software Development :: Build Tools
+    Topic :: System :: Systems Administration
+platforms = POSIX, Windows
+provides = plumbum
+
+[options]
+packages =
```
henryiii: Is there a reason not to use automatic package discovery here?

setup.py
```diff
@@ -52,37 +49,11 @@ def run(self):
...
-        "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.4",
-        "Programming Language :: Python :: 3.5",
-        "Programming Language :: Python :: 3.6",
-        "Programming Language :: Python :: 3.7",
-        "Programming Language :: Python :: 3.8",
-        "Topic :: Software Development :: Build Tools",
-        "Topic :: System :: Systems Administration",
-    ],
+    use_scm_version={
```
henryiii: I think I'd prefer the PEP 517 method of setuptools_scm; I'm pretty sure the produced wheel is still valid on older pip's, and that way on a build system you can ensure the correct versions of things (non-wheel installs would require a recent pip). Thoughts?

setup.py
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env python
 import os
+from pathlib import Path
```
henryiii: I take it that non-wheel installs would require Python 3? I think that's okay... We could make this nicer with a pyproject.toml though that imported pathlib2 for use in the setup.py on Python 2.

setup.cfg
KOLANICH: I can make it automatic. Should I?

henryiii: Sure. As a general rule it’s safer.

KOLANICH:
Travis CI build has failed because [psutil has dropped pypy 2.7](https://github.com/giampaolo/psutil/issues/1659). Taking in account that python 2.7 is EOL, should we drop pypy 2.7 too or should we just apply a patch removing the antifeature from psutil?

henryiii:
I'll try to fix the CI soon. I'm personally fine with dropping PyPy 2.7 if other libraries are dropping it.

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[build-system]
+requires = ["setuptools>=44", "wheel", "setuptools_scm[toml]>=3.4.3"]
```
henryiii:
```suggestion
requires = ["setuptools>=42", "wheel", "setuptools_scm[toml]>=3.4.3"]
```

You only need 42 or better, and 44 kills Python 3.4. Not that I'm worried about 3.4, but if it's easy to fix here, we might as well.

henryiii: You should add `pathlib;python_version<"3"` here too, since it's needed for setup.py.  This won't fix Python 2.6 CI, since 2.6 does not have a new enough pip installed to respect this file. But that's okay, we can install pathlib manually in CI.

None of this affects the wheel, which is what 100% of users install.

setup.py
```diff
@@ -52,37 +49,7 @@ def run(self):
...
-    license = "MIT",
-    url = "https://plumbum.readthedocs.io",
-    packages = ["plumbum", "plumbum.cli", "plumbum.commands", "plumbum.machines", "plumbum.path", "plumbum.fs", "plumbum.colorlib"],
-    package_data={
-        "plumbum.cli":["i18n/*/LC_MESSAGES/*.mo"]
-    },
-    platforms = ["POSIX", "Windows"],
-    provides = ["plumbum"],
-    keywords = "path, local, remote, ssh, shell, pipe, popen, process, execution, color, cli",
     cmdclass = {'test':PyTest,
```
henryiii: I'm okay with killing off 'test' and `test_requires` too, by the way (doesn't need to be in this PR); Python is deprecating `setup.py test`. The only correct way to run the tests is via `pytest`.

setup.cfg
```diff
@@ -1,5 +1,57 @@
...
+    plumbum
+    plumbum.cli
+    plumbum.commands
+    plumbum.machines
+    plumbum.path
+    plumbum.fs
+    plumbum.colorlib
+setup_requires = setuptools>=44; wheel; setuptools_scm[toml]>=3.4.3
```
henryiii: Please remove this line. `setup_requires` is deprecated.

KOLANICH: I don't think we should remove it (older versions of setuptools that don't know about pyproject.toml should understand that).

But as you wish, it's not a very strong issue, ones staying on bleeding edge (like me) should not encounter any issues from that.

pyproject.toml
KOLANICH: It turned out that we don't need pathlib anymore since we have moved the stuff about `setuptools_scm` to `pyproject.toml`.

setup.cfg
KOLANICH: It turned out without this line builds fail.

henryiii: Okay. Should only affect building wheels, which only devs and CI does, but maybe we don’t have the CI up to date. It doesn’t hurt to have it.

henryiii:
@KOLANICH, how about we merge #506 , then make a 1.6.9 release, then dump Python 2.6 for good and merge this as the beginning of 1.7.0?

KOLANICH:
Do what you feel is right. IMHO it is OK to have 10 seconds between 2 releases :).

henryiii:
1.6.9 is out. Feel free to drop 2.6!

henryiii:
Fantastic, thanks! Thanks also for cleaning out a few ugly 2.6 workarounds!

PR #500, opened 2020-02-25
fix support for win32 in iter_lines

koreno:
This feature never worked on win32, since it doesn't support using `select` with anything but sockets...
The only way I found to implement it was to use two threads


henryiii:
Can you rebase on master, please? Fixed CI.

koreno:
updated

- Ofer

henryiii:
Thanks!

PR #505, opened 2020-03-12
Fix remote stat OSError construction

### LINKED ISSUE #504: Remote path stat has odd OSError
Looking at the following it seems oddly wrong:

https://github.com/tomerfiliba/plumbum/blob/master/plumbum/path/remote.py#L170-L174

```
>>> x = OSError(errno.ENOENT)
>>> x.message
2
>>> x.errno
```

Seems like message is getting populated with the errno value and not errno :-/

harlowja:
Fixes #504


henryiii:
Thanks!

harlowja:
No problem.

PR #513, opened 2020-06-16
fix handling of env-vars passed to plumbum Commands; support new with_cwd

koreno:
Fix handling of env-vars passed to plumbum BoundEnvCommands:
    - don't use the non-threadsafe and session-dependent .env() context manager
    - sync popen support for 'env' param in all machine impls: local/ssh/paramiko
    - add .with_cwd() to complement .with_env()

### DIFF
--- a/docs/_cheatsheet.rst
+++ b/docs/_cheatsheet.rst
@@ -21,6 +21,15 @@ also :ref:`import commands <import-hack>`:
     >>> grep
     LocalCommand(<LocalPath /bin/grep>)
 
+Or, use the ``local.cmd`` syntactic-sugar:
+
+.. code-block:: python
+
+    >>> local.cmd.ls
+    LocalCommand(<LocalPath /bin/ls>)
+    >>> local.cmd.ls()
+    u'build.py\ndist\ndocs\nLICENSE\nplumbum\nREADME.rst\nsetup.py\ntests\ntodo.txt\n'
+
 See :ref:`guide-local-commands`.
 
 Piping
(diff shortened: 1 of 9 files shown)


AndydeCleyre:
Is there more documentation for `with_cwd`? How does it relate to the usual `local.cwd`? Would the following be valid?

```python
from plumbum import local
from plumbum.cmd import pwd

with local.cwd('/tmp'):
    pwd.with_cwd('/home')().strip() == '/home'
    pwd().strip() == '/tmp'
```

henryiii:
Could you please address @AndydeCleyre's question? I'm a little worried if we are just adding another way to spell something that already exists, that's just more to maintain and adapt for Plumbum 2.0.

koreno:
> Is there more documentation for `with_cwd`? How does it relate to the usual `local.cwd`? Would the following be valid?
> 
> ```python
> from plumbum import local
> from plumbum.cmd import pwd
> 
> with local.cwd('/tmp'):
>     pwd.with_cwd('/home')().strip() == '/home'
>     pwd().strip() == '/tmp'
> ```

sorry it took so long -
I would expect the `with_cwd` to override the ctxmgr, since it is set directly on the command object. Only command objects that don't have their own cwd will fall back on the default, which is modified by the ctxmgr.
I will add a test for this, and docs.

What do you think?

henryiii:
With test and docs, I'd be fine with it.

koreno:
@henryiii  - ci is failing on coverall actions, not sure how to proceed...

henryiii:
Looks like it is due to this: https://github.com/coverallsapp/github-action/issues/7

We'll probably have to disable coverage on PRs that are forked, unless there's some other way around this. `pull_request_target` is a new option, but it runs on the base and not the merged state, so can't be used for coverage.

henryiii:
Oh, that's odd, this is not a fork. Should be getting a read/write token. (Probably still need to fix this for forks, but not as sure this is the problem).

henryiii:
Coveralls 3 broke GitHub service detection ( https://github.com/TheKevJames/coveralls-python/issues/252 ), so I now manually specify it and it should be working again.

PR #515, opened 2020-09-10
ssh: better error reporting on SshSession errors

koreno:
We fail so frequently on various ssh configuration problems, and yet the error messages are mostly uninformative:

```
01:59:07 | No stderr result detected. Does the remote have Bash as the default shell?
...
01:59:07 |   */plumbum/machines/session.py:277 ................................ run >> return run_proc(self.popen(cmd), retcode)
01:59:07 |   */plumbum/commands/processes.py:292 ......................... run_proc >> stdout, stderr = proc.communicate()
01:59:07 |   */plumbum/machines/session.py:136 ........................ communicate >> msgerr) if name == "2" else SSHCommsError(msg)
```

It seems that the code was not ready for various other return codes

### DIFF
--- a/plumbum/commands/processes.py
+++ b/plumbum/commands/processes.py
@@ -127,8 +127,9 @@ class ProcessExecutionError(EnvironmentError):
     well as the command line used to create the process (``argv``)
     """
 
-    def __init__(self, argv, retcode, stdout, stderr):
+    def __init__(self, argv, retcode, stdout, stderr, message=None):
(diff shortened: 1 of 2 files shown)


henryiii:
(retriggering the CI)

henryiii:
This seems to break PyPy?

henryiii:
(I'm happy to try to move to GHA which will upgrade PyPy (this is 7.1.1-beta ?), then we can investigate. It might be that you can't do multiple inheritance on PyPy for the exceptions?

PR #498, opened 2020-02-07
Added __path__

KOLANICH:

PR #522, opened 2020-11-12
Fix TypeError:  'str' object is not callable in LocalPath.__fspath__

moskupols:
So basically there was a bug in `LocalPath.__fspath__`: we tried to call `self._path()` where `_path` is a property.
I solved the problem by just inheriting the implementation from Path, as `Path.__fspath__` is implemented correctly:
```
    def __fspath__(self):                                                                                                                          
        """Added for Python 3.6 support"""                                                                                                         
        return str(self)    
```

### DIFF
--- a/plumbum/path/local.py
+++ b/plumbum/path/local.py
@@ -58,9 +58,6 @@ class LocalPath(Path):
             cls, os.path.normpath(os.path.join(*(str(p) for p in parts))))
         return self
 
-    def __fspath__(self):
-        return self._path()
-
     @property
     def _path(self):
         return str(self)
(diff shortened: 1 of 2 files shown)


henryiii:
Assuming it passes tests, this looks fine to me. Thanks!

moskupols:
The tests failed, but I don't think it's because of my patch: seems like there's a problem with paramiko and OpenSSL 1.0.2 no longer being supported.

moskupols:
Created a separate issue for the build problem: #523

koreno (approved):

henryiii:
Now questioning to myself why we needed this on LocalPath if Path already had it? @koreno I think you added #498? Not that it hurts.

koreno:
> Now questioning to myself why we needed this on LocalPath if Path already had it? @koreno I think you added #498? Not that it hurts.

not me, @KOLANICH did. indeed it seems redundant, since it's already on the base class since 2b5a82a6

Issue #523, opened 2020-12-08
Travis pypy build fails because of stale openssl


moskupols:
This build of a branch that doesn't do anything with ssl/paramiko fails just for pypy: https://travis-ci.org/github/tomerfiliba/plumbum/builds/747879936
As far as I understand, the reason is that paramiko depends on cryptography, and release of cryptography==3.2.0 dropped support for OpenSSL 1.0.0: https://cryptography.io/en/latest/changelog.html#v3-2
At the time of the last commit to the master that version of cryptography was not released yet, so that's why the problem did not affect the status of master (yet).

I'm not sure why the only failing build is the pypy one, but it fails consistently while the other python builds don't.
Maybe travis has slightly different VM images for different python implementations?

I see a couple of possible solutions on .travis.yml side:
- Add something like `sudo apt install libssl1.1` to travis.sh
- Add something like `env: CRYPTOGRAPHY_ALLOW_OPENSSL_102=1` to pypy configuration

Unfortunately I don't have a set up to test these options, but it should suffice.

moskupols:
Oh, I guess I can just test these approaches with a new PR.

PR #524, opened 2020-12-08
Install libssl1.1 in travis.sh

moskupols:
This should address #523

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
     env: PYV=3.8
   - python: pypy3
     before_install:
-      apt-get install libssl1.1
+    - sudo apt-get install libssl1.1
   - language: generic
     env: PY3=2 PYV=Mac2
     os: osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,7 @@ matrix:
   - python: 3.8
     env: PYV=3.8
   - python: pypy3
-    before_install:
-    - sudo apt-get install libssl1.1
+    env: CRYPTOGRAPHY_ALLOW_OPENSSL_102=1
   - language: generic
     env: PY3=2 PYV=Mac2
     os: osx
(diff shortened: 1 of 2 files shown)


henryiii:
Ping me when you have this working or if you need help. We can get the other PR in as soon as this gets cleaned up.

At some point, if I ever have "free" time, I need to migrate this over to GHA.

.travis.yml
```diff
@@ -17,7 +17,7 @@ matrix:
     env: PYV=3.8
   - python: pypy3
     before_install:
-      apt-get install libssl1.1
+    - sudo apt-get install libssl1.1
```
henryiii: I'd try sudo apt-get update && sudo apt-get install libssl1.1 - often the cache is empty in systems like this.

moskupols: Sadly the problem is there's no libssl1.1 for xenial: the earliest ubuntu that has it in the standard repository is ubuntu 18 bionic.
Ubuntu 16 xenial is nearing its end of support.

moskupols:
Even if I allow cryptography to link against the old version of OpenSSL, it still fails to build because of some redefinition between its sources and the headers of libssl.
Cryptography installation notes don't mention Ubuntu 16 among supported versions: https://cryptography.io/en/latest/installation.html .
So I'm afraid in order to test pypy builds with latest cryptography it may be required to drop testing for xenial.
I guess another solution would be to put a constraint cryptography<3.2.0?

henryiii:
Then can't you bump the OS? Travis has bionic, I think? `dist: bionic`? Possibly for this one job for now?

.travis.yml
```diff
@@ -16,6 +16,8 @@ matrix:
   - python: 3.8
     env: PYV=3.8
   - python: pypy3
```
henryiii, line 18:
```suggestion
  - python: pypy3
    dist: bionic
```

IIRC this would work?

moskupols:
@henryiii Yeah looks like that solved the problem

PR #527, opened 2020-12-12
fixes RemotePath.copy()

hassium-269:
call RemotePath.delete (a method) instead of RemotePath.remove (not a method)


henryiii:
Ideally, we should add a test for this too. Thanks!

PR #539, opened 2021-04-05
Fixing project license in docs config file

tapdiego-amzn:
The documentation config file seems to identify the project license as CC-BY-SA 3.0; fixing it to say MIT instead, as that is the actual license.


henryiii:
Tests were broken by the recent move to `ubuntu-20.04` for `ubuntu-latest`, fixed after #540.

henryiii:
I'd be quite shocked if this broke a test, but green checkmarks are pretty. :)

PR #546, opened 2021-04-22
Removed setup.py.

KOLANICH:
It turned out that only setup.cfg is enough for modern setuptools.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,0 @@
-#!/usr/bin/env python3
-
-from setuptools import setup
-
-setup()


henryiii:
Yeah, but then you can't do editable installs for development 👎 , and you loose GitHub's insights for Python projects, since it doesn't trigger as a Python project (as of about a year ago, maybe they've fixed that now).

henryiii:
The first problem will be fixed in the next Pip release: https://github.com/pypa/pip/pull/9547 :) So I'd be more-or-less fine with it once there's a new pip. It does mean Python 2 and 3.5 development will be harder, but a) I don't do Python 2 development, I just fix CI issues when it fails, and b) these versions of Python will be dropped in the version after the "split-up" next version.

henryiii:
The dependency graph is broken currently anyway, so losing setup.py won't be a big deal for that.

henryiii:
Pip 21.1 is out.

piotr-dobrogost:
Are you aware of [Remove support for setup.cfg only projects](https://github.com/pypa/pip/pull/10031) change in pip?

henryiii:
That's poorly named. That only removes support for projects that have a `setup.cfg` but _not_ a `pyproject.toml`, which was accidentally added in 21.1.3. We have a `pyproject.toml`.

And, yes, I was the one who caught the problem in 21.1.3 and asked for that PR, if you follow through the various discussions. :)

henryiii:
I'm going to be _very_ tempted to move to PyPA/flit with PEP 621 configuration instead of just deleting setup.py once we drop 2.7 & 3.5 support. The only thing we'd loose is the automatic versioning, which is admittedly a bummer. Thoughts? (By the time we do it, the PR to add PEP 621 to setuptools might be ready, but flit is still 10 faster...)

KOLANICH:
>The only thing we'd loose is the automatic versioning, which is admittedly a bummer. 

If I remember right, `poetry_core` is capable to call setuptools plugins like `setuptools_scm`. Though I don't approve `poetry` tool itself (for it being adept of a cargo cult of virtualenvs), `poetry_core` doesn't follow this approach and so works pretty fine.

henryiii:
I think PyPA/Hatch's hatchling is the right direction for Plumbum to rid itself of setup.py, FYI.

henryiii:
FYI, after a little bit (to make sure there aren't regressions in 1.8.0), I'll drop 3.6 then merge the hatchling PR, which will remove setup.py. :)

KOLANICH:
OK.

PR #547, opened 2021-04-22
Allowed to request processing of all possible values in Set

KOLANICH:


henryiii:
Needs tests, Python 2 supers (for now). Also `*` seems problematic (shells are likely to expand it, requiring `'*'` to pass it in), but that's probably okay. Does this customize the help output at all? Also, why is CSV on by default? Seems like an odd departure from Set. Maybe AllCSV should be an instance of AllSet? In fact, why not use the same method used by CSV to make these, and build it in as an option to the main Set (initially empty). Then `AllSet = Set(all={"all", "*"})` and it's also easy to customize by a user.

KOLANICH:
>Also, why is CSV on by default? Seems like an odd departure from Set.

By definition `all` means all the values within a set. This implies that multiple values are expected to be passed into a single argument. But just `Set` with `csv=False` accepts only a single value; Do I understand right how ordinary `Set` is meant to work?

>In fact, why not use the same method used by CSV to make these, and build it in as an option to the main Set (initially empty). Then AllSet = Set(all={"all", "*"}) and it's also easy to customize by a user.

I have considered that, but it can either break existing apps (if we ship `all` prepopulated by default) or can cause "a zoo" (if we don't, then each dev would invent an own value for `all`, which would be a usability issue). Both cases are undesireable. So it is proposed for the ones needing multiple values in a set explicitly opt-in into `AllSet`, if they need its functionality (if they allow multiple values in it, then they almost always need it).

KOLANICH:
Integrated the functionality into `Set`, added tests and did a bit of refactoring

plumbum/cli/switches.py
```diff
@@ -464,18 +464,47 @@ class MyApp(Application):
                              comparison or not. The default is ``False``
     :param csv: splits the input as a comma-separated-value before validating and returning
                 a list. Accepts ``True``, ``False``, or a string for the separator
-    """
+    :param all_markers: when a user inputs any value from this set, it is considered that
+                        he requested iteration over all values. By default `all_markers`
```
henryiii, line 459:
```suggestion
                        they requested iteration over all values. By default `all_markers`
```

plumbum/cli/switches.py
```diff
@@ -464,18 +464,47 @@ class MyApp(Application):
...
+                        are {"*", "all"}."""
 
     def __init__(self, *values, **kwargs):
         self.case_sensitive = kwargs.pop("case_sensitive", False)
+        all_markers = kwargs.pop("all_markers", None)
+        if all_markers is None:
+            all_markers = {"*", "all"} - set(values)
```
henryiii, line 466:
```suggestion
            all_markers = {"*"} - set(values)
```

Adding the English string "all" by default seems a little likely to be suprising, and possibly conflicting with potential entries. I think it would be better to default to "*" then provide an example for how to set `{"*", "all"}`.

plumbum/cli/switches.py
```diff
@@ -464,18 +464,47 @@ class MyApp(Application):
                              comparison or not. The default is ``False``
     :param csv: splits the input as a comma-separated-value before validating and returning
                 a list. Accepts ``True``, ``False``, or a string for the separator
-    """
+    :param all_markers: when a user inputs any value from this set, it is considered that
+                        he requested iteration over all values. By default `all_markers`
+                        are {"*", "all"}."""
```
henryiii, line 460:
```suggestion
                        are {"*"}; you can set it to {"all", "*"}, for example."""
```

plumbum/cli/switches.py
KOLANICH: For "*", as you have noticed, one has to type quotes. It is not very convenient (it is the same amount of characters, but to type a `"` one has to press shift, it is also a "mode switch "from letters to punctuation"), so I usually use "all".

>Adding the English string "all" by default seems a little likely to be suprising

I don't think so: if one wants a string "all" in an arg of type "Set", it explicitly says that he wants all the values.

If you mean that the problem is that the string is English that "discriminates" against other languages, I think that having the universal same string for all the languages that doesn't depend on a locale outweights the possibility of mismatch between languages of other enum values and outweights the variant not having it at all. Making a language-dependent marker will cause only issues and increase cognitive load in using it. Not having a default one will cause users to invent own ones, which again will increase cognitive load of using it.

>, and possibly conflicting with potential entries. 

Introducing any default string can be conflicting. That's why the line to which you have left this comment exists - to avoid the conflicts automatically. If user-provided Set doesn't have `all` and `*`, then they will be used as markers for all the values. If user explicitly has opted in, then he fully controls that field and it is he is responsible for solving for possible conflicts.

plumbum/cli/switches.py
KOLANICH: https://stallman.org/articles/genderless-pronouns.html

So, in order to have `they` (if one strictly wants to avoid possibility of stirring up SJ-hornets' nests), `when a user inputs` should become `when users input`.

henryiii:
```suggestion
                        to be a request for iteration over all values. By default `all_markers`
```

plumbum/cli/switches.py
KOLANICH: Why do you suggest it be changed, if we resolve conflicts automatically?

henryiii: Plumbum comes with built-in translations for German, French, Dutch, and Russian. Either this string needs to support translations, or it should be language agnostic. A local dependent default is poor (as you noted), so why not just use the symbol? Also, if the user uses all-caps for their options, then this will be lower case, which is weird. Why not just use `"*"` as the default string (exactly the same number of characters as `all`) and let the user decide if they like "all" or something else?

Also there's the issue of being misleading. "all" is much more likely to look like a normal option that "*" is; and it's not a normal option, it just selects all the others.

I don't think anyone else has an "all/*" like option. Does anyone else? I checked Click, doesn't seem to have this sort of select-all feature.

henryiii: Does this show up in the help anywhere?

If you really want to make "all" part of the default, I'd take a tie breaker opinion. @tomerfiliba perhaps? Or any other person who has committed anything.

plumbum/cli/switches.py
henryiii: (and also remove "that" from the line above)

PR #550, opened 2021-06-30
Add support for OpenSSH forwarding Unix sockets

lbenezriravin:
I tried to make the interface as sane as possible while maintaining as a small diff and backwards compatibility. Let me know if you'd like it to look different. Thanks for working on this project!

OpenSSH clients can forward Unix sockets instead of TCP ports with the
same `-L` flag using a slightly modified argument. This patch should
allow users to take advantage of the full syntax. From the man pages:

```
     -L [bind_address:]port:host:hostport
     -L [bind_address:]port:remote_socket
     -L local_socket:host:hostport
     -L local_socket:remote_socket
```

This patch also adds a protection for the feature.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,9 @@
+1.8.0
+-----
+
+* SSHMachine: support forwarding Unix sockets in ``.tunnel()`` (`#550 <https://github.com/tomerfiliba/plumbum/pull/550>`_)
+
+
 1.7.0
 -----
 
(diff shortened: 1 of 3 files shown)


lbenezriravin:
@henryiii Is this a feature you're open to?

henryiii:
Yes, I just hadn't reviewed it yet. Could you add an entry to the changelog?

lbenezriravin:
Sure thing. Thanks for taking a look.

henryiii:
Thanks!

PR #553, opened 2021-07-14
Introducing new Application property: COLOR_GROUP_TITLES

ink-splatters:
Hello,

Thanks a lot for the great framework!

I'm migrating a CLI app from `click` to `plumbum` and it's going like a charm!

But got an idea for a small feature/improvement;

hence the pull request:

- `COLOR_GROUP_TITLES` property introduced and documented
- `test_clicolor.py` modified to cover the new case

Cheers,
Peter


henryiii (approved):
LGTM! Thanks!

PR #560, opened 2021-10-05
tests: include Python 3.10 (linux only)

henryiii:


henryiii:
Looks like we can't support Windows 3.10 yet: https://github.com/mhammond/pywin32/issues/1588

PR #562, opened 2021-10-17
Add convenience flag for reverse ssh tunnels

oakreid:
SSH supports the creation of reverse tunnels using a very similar syntax to normal tunnels. From the man pages:
```
     -R [bind_address:]port:host:hostport
     -R [bind_address:]port:local_socket
     -R remote_socket:host:hostport
     -R remote_socket:local_socket
     -R [bind_address:]port
```
This PR adds a flag to the `.tunnel()` method of `SshMachine` that allows for easy creation of a reversed SSH tunnel

### DIFF
--- a/tests/test_remote.py
+++ b/tests/test_remote.py
@@ -5,6 +5,7 @@ import socket
 import sys
 import time
 from copy import deepcopy
+from multiprocessing import Process, Queue
 
 import env
 import pytest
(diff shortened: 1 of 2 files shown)


henryiii:
Thanks! Didn't format or catch that time_ns was included, but doing that in #566.

tests/test_remote.py
```diff
@@ -477,6 +478,46 @@ def test_tunnel(self):
                 print(p.communicate())
                 assert data == b"hello world"
 
+
+    def test_reverse_tunnel(self):
+
+        def serve_reverse_tunnel(queue):
```
henryiii, line 484: FYI, this is not picklable, so can't be forked. Why did this only matter on Python 3.10 + macOS? I have no idea. Anyway, moving it out of the test function should fix the problem.

oakreid: That's very strange. Thanks for cleaning that up!

henryiii: Think you could look into possibly making the test support forked multiprocessing? Not a big issue if not (the MP part is really due to the test needing both sides of the tunnel, not a normal occurrence).

oakreid: Yeah I can definitely look into it!

PR #563, opened 2021-10-25
setDaemon is deprecated in python 3.10

timgates42:
Setting the python property is supported python 2.6+


henryiii:
Thanks! And we are 2.7+ now. :) (3.6+ Jan 2022)

PR #566, opened 2021-11-22
fix: issue in recent PR plus more testing

henryiii:
Fix for #562, plus testing improvements and skips. More coverage (full 3.10, PyPy 3.8).

### DIFF
--- a/tests/test_remote.py
+++ b/tests/test_remote.py
@@ -478,9 +478,7 @@ class TestRemoteMachine(BaseRemoteMachineTest):
                 print(p.communicate())
                 assert data == b"hello world"
 
-
     def test_reverse_tunnel(self):
-
         def serve_reverse_tunnel(queue):
             s = socket.socket()
             s.bind(("", 12222))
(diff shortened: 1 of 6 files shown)


tests/test_remote.py
```diff
@@ -505,19 +506,20 @@ def serve_reverse_tunnel(queue):
...
 s = socket.socket()
 s.connect(("localhost", {}))
 s.send("{}".encode("ascii"))
 s.close()
-""".format(remote_socket, message)
+""".format(
+                    remote_socket, message
+                )
                 (rem.python["-u"] << remote_send_af_inet).popen()
                 tunnel_server.join()
```
henryiii, line 524: @oakreid Now this just hangs on macOS Python 3.10. Any idea here? It's probably using a different mp model?

henryiii: Ah, this only works when forking, not when spawning, and the default changed in 3.8+. Skipping on macOS 3.8+ should be fine for now. Ideally, it would be nice to support spawn too.

PR #568, opened 2021-11-23
Use thread instead of multiprocessing in reverse ssh tunnel unit test

oakreid:


henryiii:
(PS: pypy is flaky still, don't worry about pypy)

henryiii (approved):

PR #570, opened 2021-12-01
Directly reference pywin32 in requirements 

silasary:
`pypiwin32` is an unmaintained redirect to `pywin32`

This PR changes the requires to point directly at the `pywin32` package

It also means that pywin32 will correctly have the `platform_system=='Windows'` markers, which is important for projects using Pipenv


henryiii:
Thanks!

PR #579, opened 2022-01-28
fix: add types and fix issue with ioctrl + Python 3

henryiii:
Started adding types (using monkeytype), and hit a possible bug that was probably hidden by a too-broad except.

### DIFF
--- a/plumbum/cli/termsize.py
+++ b/plumbum/cli/termsize.py
@@ -7,11 +7,12 @@ import os
 import platform
 import warnings
 from struct import Struct
+from typing import Optional, Tuple
 
 from plumbum import local
 
 
-def get_terminal_size(default=(80, 25)):
+def get_terminal_size(default: Tuple[int, int] = (80, 25)) -> Tuple[int, int]:
     """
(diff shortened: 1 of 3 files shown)


plumbum/cli/termsize.py
```diff
@@ -71,18 +72,19 @@ def _get_terminal_size_tput():  # pragma: no cover
...
-def _ioctl_GWINSZ(fd):
+def _ioctl_GWINSZ(fd: int) -> Optional[Tuple[int, int]]:
     yx = Struct("hh")
     try:
         import fcntl
         import termios
 
-        return yx.unpack(fcntl.ioctl(fd, termios.TIOCGWINSZ, "1234"))
+        # TODO: Clean this up. Problems could be hidden by the broad except.
+        return yx.unpack(fcntl.ioctl(fd, termios.TIOCGWINSZ, b"1234"))  # type: ignore[return-value]
```
henryiii, line 82: This was the possible bug, it accepts a byte string in Python 3, not a unicode string.

PR #582, opened 2022-01-28
iter_lines: added new 'buffer_size' parameter, and updated docstrings

koreno:


plumbum/commands/processes.py
```diff
@@ -304,6 +303,7 @@ def run_proc(proc, retcode, timeout=None):
 BY_POSITION = object()
 BY_TYPE = object()
 DEFAULT_ITER_LINES_MODE = BY_POSITION
+DEFAULT_BUFFER_SIZE = _INFINITE = float("inf")
```
henryiii, line 306:
```suggestion
DEFAULT_BUFFER_SIZE = float("inf")
```

Don't think there's a point to keeping this around.

henryiii (approved):
Seems fine. I think we should try to look as much like Python 3's subprocess, but this method is not one from subprocess AFAIK, so don't have strong opinions on adding a bit of functionality here. If my one tiny suggestion is fine, I think it's good to merge.

plumbum/commands/processes.py
```diff
@@ -335,25 +336,41 @@ def iter_lines(
...
     for t, line in _iter_lines(proc, decode, linesize, line_timeout):
 
         # verify that the proc hasn't timed out yet
         proc.verify(timeout=timeout, retcode=None, stdout=None, stderr=None)
 
-        buffers[t].write(line + "\n")
+        buffer = buffers[t]
+        if buffer_size > 0:
+            buffer.append(line)
+            if buffer_size < _INFINITE:
```
henryiii, line 372: Oh, this is why, sorry.

PR #583, opened 2022-01-28
feat: cache commands and remember host in exceptions

koreno:
I _could_ break it apart into individual pull-requests, but thought you might be okay with merging it as is more or less, and then refining it where needed?


plumbum/commands/base.py
```diff
@@ -335,7 +335,7 @@ def popen(self, args=(), **kwargs):
 
 
 class BoundEnvCommand(BaseCommand):
-    __slots__ = ("cmd",)
+    __slots__ = ("cmd", "env", "cwd")
```
henryiii: Merge mistake? "env" and "cwd" are already in `BaseCommand` `__slots__`.

plumbum/machines/remote.py
```diff
@@ -173,6 +173,7 @@ def __init__(self, encoding="utf8", connect_timeout=10, new_session=False):
         self.uname = self._get_uname()
         self.env = RemoteEnv(self)
         self._python = None
+        self._program_cache = {}
```
henryiii, line 176: I thought through ways to be more clever here with functools.lru_cache, but they end up being worse (it caches all arguments, including `self`). So I think this is good.

plumbum/machines/session.py
```diff
@@ -72,7 +72,8 @@ class SessionPopen(PopenAddons):
     """A shell-session-based ``Popen``-like object (has the following attributes: ``stdin``,
     ``stdout``, ``stderr``, ``returncode``)"""
 
-    def __init__(self, proc, argv, isatty, stdin, stdout, stderr, encoding):
+    def __init__(self, host, proc, argv, isatty, stdin, stdout, stderr, encoding):
```
henryiii: Wouldn't it be nicer to add `host` to the end, rather than the beginning, so as to not shift all the arguments? I'd be fine to make it keyword only, too. ;)

```suggestion
    def __init__(self, proc, argv, isatty, stdin, stdout, stderr, encoding, *, host):
```

plumbum/machines/session.py
```diff
@@ -292,6 +301,7 @@ def popen(self, cmd):
         self.proc.stdin.write(full_cmd + b"\n")
         self.proc.stdin.flush()
         self._current = SessionPopen(
+            self.host,
```
henryiii: See previous comment, this would be moved.

plumbum/machines/session.py
```diff
@@ -202,7 +208,10 @@ class ShellSession:
                             is seen, the shell process is killed
     """
 
-    def __init__(self, proc, encoding="auto", isatty=False, connect_timeout=5):
+    def __init__(
+        self, proc, encoding="auto", isatty=False, connect_timeout=5, host=None
```
henryiii:
```suggestion
        self, proc, encoding="auto", isatty=False, connect_timeout=5, *, host=None
```

It's harder to change existing args, but new args should be kw-only.

plumbum/commands/base.py
koreno: probably.
won't this override? or you're saying each base class keeps its own `__slots__`?

plumbum/commands/base.py
```diff
@@ -335,8 +335,6 @@ def popen(self, args=(), **kwargs):
 
 
 class BoundEnvCommand(BaseCommand):
-    __slots__ = ("cmd",)
-
```
henryiii: This does need `__slots__ = ("cmd",)`, that's correct. It just doesn't need to redefine slots that are already present in the base class.

plumbum/commands/processes.py
```diff
@@ -113,9 +113,10 @@ class ProcessExecutionError(EnvironmentError):
     well as the command line used to create the process (``argv``)
     """
 
-    def __init__(self, argv, retcode, stdout, stderr, message=None):
+    def __init__(self, argv, retcode, stdout, stderr, message=None, host=None):
```
henryiii:
```suggestion
    def __init__(self, argv, retcode, stdout, stderr, message=None, *, host=None):
```

One more I missed.

PR #584, opened 2022-01-29
ci: nicer pylint annotations

henryiii:
Also cleaning up the double assignment and `_INFINITY` constant from  #582 using `math.inf` from Python 3.5+.

- ci: nicer pylint output
- style: use math.inf

### DIFF
--- a/.github/matchers/pylint.json
+++ b/.github/matchers/pylint.json
@@ -1,0 +1,32 @@
+{
+  "problemMatcher": [
+    {
+      "severity": "warning",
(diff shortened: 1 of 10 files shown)


henryiii:
Ahh, due to the fact we haven't fixed some of the remaining warnings, this will be too noisy if we turn on annotations now. At least we can enjoy the nice annotations example here for now. :)

henryiii:
Fixing some of them, and TODO'ing the rest.

PR #585, opened 2022-01-31
Use f-strings instead format.

damani42:
In the documentation use f-strings in the examples.


henryiii:
Nice, forgot about docs. Automated tools and pylint checks don't cover docs. :)

damani42:
You're welcome @henryiii and I prepare another patch about something else.

PR #586, opened 2022-02-02
exceptions: ensure ProcessExecutionError can be pickled

koreno:
Something in recent refactors broke this capability, which we are using in our infrastructure, so I've fixed and cemented it with a UT.

### DIFF
--- a/plumbum/commands/processes.py
+++ b/plumbum/commands/processes.py
@@ -115,7 +115,7 @@ class ProcessExecutionError(OSError):
 
     def __init__(self, argv, retcode, stdout, stderr, message=None):
 
-        # we can't use 'super' here since EnvironmentError only keeps the first 2 args,
+        # we can't use 'super' here since OSError only keeps the first 2 args,
         # which leads to failuring in loading this object from a pickle.dumps.
         Exception.__init__(self, argv, retcode, stdout, stderr)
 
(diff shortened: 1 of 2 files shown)


henryiii:
EnvironmentError isn't a real error anymore in Python 3, it's just an OSError. OSErrors return subclasses based on what the OS return code in the first argument is. I think we need to refactor this a bit. Specifically, the first value is the return code, and for us it's the second, and OSError's new is where the subclass selection is happening - this sounds very incorrect to me! For now, this is fine to fix the immediate problem.

henryiii:
Maybe it is implemented in `__init__`, since this seems to be stable (after the patch), or maybe it checks the cls before doing the subclass (though `EnvironementError(3, "msg")` returns `ProcessLookupError(3, 'msg')`, etc). The exceptions are mostly implemented in C, so hard to quickly check.

plumbum/commands/processes.py
```diff
@@ -114,7 +114,11 @@ class ProcessExecutionError(EnvironmentError):
     """
 
     def __init__(self, argv, retcode, stdout, stderr, message=None):
-        super().__init__(self, argv, retcode, stdout, stderr)
+
+        # we can't use 'super' here since EnvironmentError only keeps the first 2 args,
```
henryiii:
```suggestion
        # we can't use 'super' here since OSError only keeps the first 2 args,
```

henryiii:
Thanks!!!

PR #590, opened 2022-02-09
DO NOT MERGE: trying setuptools PR

henryiii:
- chore: delete setup.py
- DO NOT MERGE: trying PEP 621 directly
- DO NOT MERGE: try full PEP 621


pyproject.toml
```diff
@@ -1,11 +1,79 @@
...
     "setuptools_scm[toml]>=3.4.3"
 ]
 build-backend = "setuptools.build_meta"
 
+
+[project]
+name = "plumbum"
+description = "Plumbum: shell combinators library"
+readme = "README.rst"
+author = { name="Tomer Filiba", email="tomerfiliba@gmail.com" }
```
abravalheri: I think the standard mandates an array here (and also `authors` instead of `author`):

```suggestion
authors = [{ name="Tomer Filiba", email="tomerfiliba@gmail.com" }]
```

pyproject.toml
```diff
@@ -1,11 +1,79 @@
...
+license = { file="LICENSE" }
+requires-python = ">=3.6"
+dynamic = ["version"]
+requires = [
```
abravalheri: PEP 621 seems to use `dependencies` instead of `requires`.

```suggestion
dependencies = [
```

abravalheri (commented):
Hi @henryiii, thank you very much for testing this out!

I admit the errors are far from understandable... I have to review what is going on. Ideally the validation phase should tell you how to change `pyproject.toml` accordingly. (I have to revisit https://github.com/abravalheri/validate-pyproject, I suspect I forgot to forbid non-standardised keys to the `project` table, as was the case here with `author` and `requires`).

Could you try rebuilding with my suggestions? I am curious to see if the tests pass...

Something that might be interesting to you in future experiments: https://github.com/abravalheri/ini2toml can help to bootstrap `pyproject.toml` from `setup.cfg`

henryiii:
Thanks, that seems to be it! It was being very picky about other things, so I assumed it would be catching things like that, so didn't check as carefully. It does check for extra keys in the `tool.setuptools` section, which is _very_ helpful.

> Something that might be interesting to you in future experiments: https://github.com/abravalheri/ini2toml

Yes, I noticed that after I did this, but this was a good experience (and looks like it might have been helpful!)

henryiii:
Might be expected, but setuptools_scm is not working, `0.1.dev1` is the version.

abravalheri:
> looks like it might have been helpful!

Definitely! It helped to find a problem in `validate-pyproject`, thanks ❤️ 

> Might be expected, but setuptools_scm is not working, `0.1.dev1` is the version.

I will try to have a look on this later this weekend, thank you very much for pointing it out. In theory it should be no different. I have to double check if all the plugin hooks in setuptools are being called...

henryiii:
I'm looking at https://github.com/tomerfiliba/plumbum/runs/5141354158?check_suite_focus=true

Ahhh! That version number is broken on master. Sorry for the noise! I hate "try not to break things" defaults...

abravalheri:
> Ahhh! That version number is broken on master.

Perfect, thank you very much for the feedback!
Glad to see I didn't break a lot of stuff in my PR 😅

henryiii:
Yeah, fix in #591 - checkout@v2 does shallow clones, unlike v1, and this seems to have been using v2 but assuming v1. It was probably not noticed because releases are on tags, and setuptools_scm is happy then. I'm guessing running this in `act` would have also had the shallow clones.

> Glad to see I didn't break a lot of stuff in my PR

Well, there's still the pybind11 one. ;) There I'm working on cleaning up the Python 3 drop PR, then rebasing the attempt on that.

abravalheri:
> Well, there's still the pybind11 one. ;) There I'm working on cleaning up the Python 3 drop PR, then rebasing the attempt on that.

I promise I will have a look on that soon 😄 
Thank you very much for the help in testing it!

pyproject.toml
```diff
@@ -1,11 +1,79 @@
 [build-system]
 requires = [
-    "setuptools>=42",
+    "setuptools @ git+https://github.com/abravalheri/setuptools@support-pyproject",
```
abravalheri: @henryiii, you can also try `git+https://github.com/pypa/setuptools@experimental/support-pyproject` now.

However this branch also adds automatic discovery for packages (which supports `src-layout` without config) and considers implicit namespaces by default (PEP 420). So the `experiments` folder might end-up being included in the wheel (I suspect that previously it was not being included because setuptools predates PEP 420).

abravalheri: The exclusion rules are documented here: https://setuptools--2894.org.readthedocs.build/en/2894/userguide/package_discovery.html#automatic-discovery

abravalheri:
Thank you very much @henryiii for updating the PR.

I think the problems in the CI are probably related to the fact setuptools supports Python >= 3.7 (and the vendored version of `tomli` uses features that don't exist in 3.6... The PyPy error seems a bit flaky isn't it?

I just pushed some changes to the experimental branch related to the feedback I got from Python's Discourse. Hopefully everything will keep working :)

PR #605, opened 2022-06-22
made StdinDataRedirection compatible with modifiers writing to stdout

### LINKED ISSUE #604: StdinDataRedirection commands fail upon modification (TEE, FG, etc.)
Commands to which standard input has been supplied from a Python string – which are `StdinDataRedirection` commands – erroneously throw exception `RedirectionError` when combined with built-in modifiers –

    stdin is already redirected

– even though `stdin` is only being redirected the one time.

---

```python
>>> from plumbum import TEE, FG, local                                          

# this works

>>> local['ls']['-d', '/'] & FG                                                                                                                               
/                                                                              
>>> local['ls']['-d', '/'] & TEE                                                                                                                              
/                                     
(0, '/\n', '')

# this does not work

>>> (local['cat']['-'] << 'this breaks') & FG                                   
    …
    raise RedirectionError("stdin is already redirected")
plumbum.commands.base.RedirectionError: stdin is already redirected
```

---

Affected modifiers invoke `cmd.run(stdin=None, …)`. It appears sufficient to change these to simply **omit the keyword** pair `stdin=None`.

However, it's unclear whether this would have unwanted side-effects; and, it seems rather that `StdinDataRedirection` is in the wrong, in throwing an exception over the value `None`.

`BaseRedirection.popen` permits either `PIPE` or `None`:

```python
if self.KWARG in kwargs and kwargs[self.KWARG] not in (PIPE, None):
    raise RedirectionError(…)
```

`StdinDataRedirection.popen` only permits `PIPE`:

```python
if "stdin" in kwargs and kwargs["stdin"] != PIPE:
    raise RedirectionError(…)
```

→ As such, the resolution is perhaps that the above check in `StdinDataRedirection.popen` be made to mirror `BaseRedirection.popen`.

jesteria:
_I.e._ fixed:

    (cat << "meow") & FG

Modifiers writing to stdout – _e.g._: `FG`, `TEE`, `RETCODE(FG=True)` and
`TF(FG=True)` – invoke commands with the specification `stdin=None`.
This is handled correctly by `BaseCommand`; however, `StdinDataRedirection`
misinterprets this as a further attempt to specify stdin (_i.e._ to
specify _something_ rather than _nothing_).

This change brings `StdinDataRedirection` in line with its base class,
ignoring `None` as well as `PIPE`.

(Though a technical detail, the class would also have failed any
invocation setting `stdin=PIPE`, along the lines of: "TypeError: got
multiple values for keyword argument 'stdin'". The kwargs passed to the
wrapped command's `popen` are corrected as well.)

resolves #604

### DIFF
--- a/plumbum/commands/base.py
+++ b/plumbum/commands/base.py
@@ -547,7 +547,7 @@ class StdinDataRedirection(BaseCommand):
         return self.cmd.machine
 
     def popen(self, args=(), **kwargs):
-        if "stdin" in kwargs and kwargs["stdin"] != PIPE:
+        if kwargs.get('stdin') not in (PIPE, None):
             raise RedirectionError("stdin is already redirected")
         data = self.data
         if isinstance(data, str) and self._get_encoding() is not None:
@@ -558,8 +558,9 @@ class StdinDataRedirection(BaseCommand):
             f.write(chunk)
             data = data[self.CHUNK_SIZE :]
         f.seek(0)
+        kwargs['stdin'] = f
         # try:
-        return self.cmd.popen(args, stdin=f, **kwargs)
+        return self.cmd.popen(args, **kwargs)
         # finally:
         #    f.close()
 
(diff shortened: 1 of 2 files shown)


henryiii:
Thanks!!!

PR #619, opened 2022-10-05
feat: add all_markers to Set

henryiii:
Closes #547. Has fewer changes, does some code modernization, and is always opt-in, which gives better backward compatibility.

### DIFF
--- a/plumbum/cli/switches.py
+++ b/plumbum/cli/switches.py
@@ -1,6 +1,8 @@
+import collections.abc
 import contextlib
 import inspect
 from abc import ABC, abstractmethod
+from typing import Callable, Generator, List, Union
 
 from plumbum import local
 from plumbum.cli.i18n import get_translation_for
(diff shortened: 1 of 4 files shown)


plumbum/cli/switches.py
```diff
@@ -456,41 +458,62 @@ class MyApp(Application):
...
     def __repr__(self):
         items = ", ".join(v if isinstance(v, str) else v.__name__ for v in self.values)
         return f"{{{items}}}"
 
-    def __call__(self, value, check_csv=True):
+    def _call_iter(
+        self, value: str, check_csv: bool = True
+    ) -> Generator[str, None, None]:
         if self.csv and check_csv:
-            return [self(v.strip(), check_csv=False) for v in value.split(",")]
```
henryiii, line 478: This was a bugfix - Setting anything other than `,` was unsupported.

PR #620, opened 2022-10-05
ci: add testing for 3.11

henryiii:
Adding 3.11 testing before a release.

### DIFF
--- a/tests/test_local.py
+++ b/tests/test_local.py
@@ -807,7 +807,7 @@ class TestLocalMachine:
 
     def test_pgrep(self):
         print(local.list_processes())
-        assert list(local.pgrep("python3"))
+        assert list(local.pgrep("python"))
 
     def _generate_sigint(self):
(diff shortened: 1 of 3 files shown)


tests/test_local.py
```diff
@@ -806,7 +806,8 @@ def test_list_processes(self):
         assert list(local.list_processes())
 
     def test_pgrep(self):
-        assert list(local.pgrep("python"))
+        print(local.list_processes())
+        assert list(local.pgrep("python3"))
```
henryiii:
```suggestion
        assert list(local.pgrep("python"))
```

tests/test_local.py
```diff
@@ -806,6 +806,7 @@ def test_list_processes(self):
         assert list(local.list_processes())
 
     def test_pgrep(self):
+        print(local.list_processes())
```
henryiii:
```suggestion
        print(*local.list_processes())
```

PR #627, opened 2022-12-12
local: accept path-like (`pathlib.Path`) objects

doronz88:
This will allow users to use the API as follows:

```python
from plumbum import local
from pathlib import Path

local[Path('/path/to/binary')]()
```

### DIFF
--- a/plumbum/machines/local.py
+++ b/plumbum/machines/local.py
@@ -233,6 +233,8 @@ class LocalMachine(BaseMachine):
             return LocalCommand(cmd)
 
         if not isinstance(cmd, RemotePath):
+            # handle "path-like" (pathlib.Path) objects
+            cmd = os.fspath(cmd)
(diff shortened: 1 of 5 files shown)


plumbum/machines/local.py
```diff
@@ -233,6 +233,8 @@ def __getitem__(self, cmd):
             return LocalCommand(cmd)
 
         if not isinstance(cmd, RemotePath):
+            # handle "path-like" (pathlib.Path) objects
+            cmd = str(cmd)
```
henryiii:
```suggestion
            cmd = os.fspath(cmd)
```

Use `os.fspath`. This would support custom objects that have differing `__str__` and `__fspath__`. And it would not try to apply on things that are not path-like.

doronz88: nice one :)
done

henryiii:
Thanks!

Issue #398, opened 2018-08-01
cannot cast to string a StdinDataRedirection-bound LocalCommand


jesteria:
Having directed string input into the standard input of a `LocalCommand` (via `<<` *i.e.* `__lshift__`), I am unable to cast the resulting `StdinDataRedirection` object to string. Attempting to do so raises an uncaught exception.

    $ ipython
    Python 3.6.3 (default, Dec 20 2017, 11:51:43) 
    Type 'copyright', 'credits' or 'license' for more information
    IPython 6.2.1 -- An enhanced Interactive Python. Type '?' for help.
    
    In [1]: import plumbum
    
    In [2]: plumbum.__version__
    Out[2]: (1, 6, 4)
    
    In [3]: cat = plumbum.local['cat']
    
    In [4]: stdin_bound_cat = cat << 'hello world'
    
    In [5]: str(stdin_bound_cat)
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    <ipython-input-37-15af60a0e450> in <module>()
    ----> 1 str(stdin_bound_cat)
    
    ~/.pyenv/versions/3.6.3/envs/test/lib/python3.6/site-packages/plumbum/commands/base.py in __str__(self)
         49
         50     def __str__(self):
    ---> 51         return " ".join(self.formulate())
         52
         53     def __or__(self, other):
    
    TypeError: sequence item 2: expected str instance, list found

Underlying this issue is that the `formulate()` method for this object returns a list which itself contains both strings and lists:

    In [6]: stdin_bound_cat.formulate()
    Out[6]: ["echo 'hello world'", '|', ['/bin/cat']]

which result cannot simply be joined by a string, as its `__str__` attempts to do.

Needless to say, this issue prevents passing such objects to `print()`, and minorly complicates introspection.

PR #629, opened 2023-01-11
Fix StdinDataRedirection's formulate() method

nebbish:
### return flat list from derived object's formulate()

The StdinDataRedirection command object was returning a list inside a list, which caused trouble when getting the "str()" of the command.

Simple example to demonstrate the problem:

    Python 3.10.8 (tags/v3.10.8:aaaf517, Oct 11 2022, 16:50:30) [MSC v.1933 64 bit (AMD64)] on win32
    Type "help", "copyright", "credits" or "license" for more information.
    >>> from plumbum.cmd import head
    >>> cmd = head['-2'] << "one\ntwo\nthree\nfour"
    >>> print(f"{cmd}")
      File "C:\Python310\lib\site-packages\plumbum\commands\base.py", line 66, in __str__
        return " ".join(self.formulate())
    TypeError: sequence item 2: expected str instance, list found
    >>>

This addresses open issue #398

### DIFF
--- a/plumbum/commands/base.py
+++ b/plumbum/commands/base.py
@@ -536,9 +536,8 @@ class StdinDataRedirection(BaseCommand):
     def formulate(self, level=0, args=()):
         return [
             f"echo {shquote(self.data)}",
-            "|",
-            self.cmd.formulate(level + 1, args),
-        ]
+            "|"
+        ] + self.cmd.formulate(level + 1, args),
 
     @property
     def machine(self):


plumbum/commands/base.py
```diff
@@ -536,9 +536,8 @@ def _get_encoding(self):
     def formulate(self, level=0, args=()):
         return [
             f"echo {shquote(self.data)}",
-            "|",
-            self.cmd.formulate(level + 1, args),
-        ]
+            "|"
+        ] + self.cmd.formulate(level + 1, args),
```
henryiii, line 540:
```suggestion
            "|",
            *self.cmd.formulate(level + 1, args),
        ]
```

Issue #548, opened 2021-06-15
Pipe has problems with large content


mtojek:
Consider the following case (bioinformatics):

`bwa mem ... | samtools view .... > outfile`, which should end up with approx. 50 GB out file. Unfortunately the pipeline hangs at ~14 GB, `bwa mem` waits on buffers and so does `samtools view`. 

Were there any reports around large file processing?

PR #632, opened 2023-01-18
Fix stalling in Pipeline command

astaric:
When a command (other than the first) in a pipeline wrote more than 64k to the stderr, and the output was consumed with the iter_lines function, the whole pipeline stalled. Fixed by reading the output of all commands where either stdout or stderr was set to PIPE.

This is the same issue as described in
https://github.com/tomerfiliba/plumbum/issues/548

Same problem still exists if command is executed with run(), but that code calls proc.communicate to read stderr/stdout and I do not know how to fix that.

### DIFF
--- a/tests/test_pipelines.py
+++ b/tests/test_pipelines.py
@@ -40,25 +40,29 @@ def test_draining_stderr_with_stdout_redirect(tmp_path, generate_cmd, process_cm
 @pytest.fixture()
 def generate_cmd(tmp_path):
     generate = tmp_path / "generate.py"
-    generate.write_text("""\
+    generate.write_text(
+        """\
 import sys
 for i in range(5000):
     print("generated", i, file=sys.stderr)
     print(i)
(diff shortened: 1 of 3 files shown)


tests/test_pipelines.py
```diff
@@ -0,0 +1,88 @@
+from typing import List, Tuple
+
+import env
```
henryiii:
```suggestion
```

vient:
@henryiii do you plan to release new version soon? Stalling pipelines is something we've bumped in several times, this fix seems important.

I don't know how new releases are made, but if there is anything I can help with, I am ready :)

henryiii:
Yeah, I'm planning one pretty soon. I want to get a release out then drop Python 3.6.

desertfury:
Hi @henryiii !
I wonder have you shipped release that contains this particular fix yet? Thanks!

PR #646, opened 2023-05-13
Add type hints to abstractmethods of Path

moskupols:
We are currently integrating pyright checks in our project. We found that generally pyright is great at inferring the return types by just looking at the source code of libraries. However for abstractmethods which only contain a docstring it assumes that the method is supposed to return None since there's no return statement.
This minimal set of hints seems to help pyright figure out how plumbum paths work.


henryiii:
Thanks! Eventually I'd like to get more typing in. :)

PR #661, opened 2023-10-31
Fix missing non-zero return code from TEE'd pipe

### LINKED ISSUE #660: TEE return code for pipeline is wrong when non-last command fails
Based on #145 
```py
from plumbum.cmd import yes, head
from plumbum import TEE, RETCODE

chain = yes['-x'] | head
print(chain & RETCODE)  # prints 1
with chain.bgrun(retcode=None) as p:
    print(p.run()[0])   # prints 1
print((chain & TEE(retcode=None))[0])  # prints 0
```

vient:
I checked that modified test is failing, and fix in this branch makes it green.

Fixes #660

### DIFF
--- a/plumbum/commands/modifiers.py
+++ b/plumbum/commands/modifiers.py
@@ -226,6 +226,7 @@ class _TEE(ExecutionModifier):
 
                         buf.append(data)
 
+            p.wait()  # To get return code in p
             stdout = "".join([x.decode("utf-8") for x in outbuf])
             stderr = "".join([x.decode("utf-8") for x in errbuf])
             return p.returncode, stdout, stderr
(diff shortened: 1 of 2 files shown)


vient:
Wat

Looks like a bug in CI, I've run nox locally, it was successful.

henryiii:
Yeah, I think it is. Will look at it when on a computer.

vient:
Btw, nox is great. I did not know what it is, installed it, and it just worked from the first try, and ran all tests? Super cool

henryiii:
See https://learn.scientific-python.org/development/guides/tasks/ - I love it. :)

I see what's wrong - we've pinned an old pylint, and it doesn't support Python 3.12, which is what the default is now. I'm fixing that in a separate PR.

PR #693, opened 2024-10-05
chore: update noxfile a bit

henryiii:
Also seeing if caching is working.


henryiii:
Yes, caching is working!

PR #700, opened 2025-03-13
Drop some unused variables

bobwhitelock:
These are no longer used since https://github.com/tomerfiliba/plumbum/pull/458.


henryiii:
CI should be fixed now.

PR #705, opened 2025-05-15
feat: add Python π, drop Python 3.8

henryiii:
I expect issues with Windows until pywin32 is ready.


henryiii:
Fix for https://github.com/python/cpython/issues/125974 applied.

PR #712, opened 2025-09-18
BaseRemoteMachine.tempdir(): add -t to mktemp to work in R/O $HOMEs

inaky-intc:
The call to mktemp with a relative template name creates the temp directory wherever the thing logs in to, $HOME.

If $HOME is read-only, then this fails.

With this change, the temp directory is made in whatever the system considers the temporary directory to be, acknowledging TMPDIR if set (per mktemp man page) and defaulting to /tmp.

### DIFF
--- a/plumbum/machines/remote.py
+++ b/plumbum/machines/remote.py
@@ -328,7 +328,9 @@ class BaseRemoteMachine(BaseMachine):
         """A context manager that creates a remote temporary directory, which is removed when
         the context exits"""
         _, out, _ = self._session.run(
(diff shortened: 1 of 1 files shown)


henryiii:
That's not what I see for `-t`:

```console
$ man mktemp
...
     -t prefix
             Generate a template (using the supplied prefix and TMPDIR if set) to create a filename template.
...
```

henryiii:
I'm on macOS, so a BSD flavor. I do see the `-p` option, and it seems consistent.

Issue #719, opened 2025-10-25
With Python 3.9, `from __future__ import annotations` breaks use of `ExistingFile` in `main` annotation


AndydeCleyre:
```console
$ python -V
Python 3.9.23
$ pip list | grep plum
plumbum 1.9.0
$ cat plumnot.py
#!/usr/bin/env python3
from __future__ import annotations

from plumbum.cli import Application, ExistingFile


class App(Application):
    def main(self, file: ExistingFile):
        print(f"{file=}")


if __name__ == "__main__":
    App()
$ ./plumnot.py plumnot.py
Traceback (most recent call last):
  File "/home/andy/Code/plumvoke/./plumnot.py", line 13, in <module>
    App()
  File "/.../plumbum/cli/application.py", line 179, in __new__
    return cls.run()
  File "/.../plumbum/cli/application.py", line 617, in run
    ordered, tailargs = inst._validate_args(swfuncs, tailargs)
  File "/.../plumbum/cli/application.py", line 554, in _validate_args
    annotation = eval(annotation)
  File "<string>", line 1, in <module>
NameError: name 'ExistingFile' is not defined
```

PR #723, opened 2025-10-29
fix: better support for future annotations on 3.9

henryiii:
Fix https://github.com/tomerfiliba/plumbum/issues/719.


AndydeCleyre:
Thanks for this run of fixes!

PR #762, opened 2025-12-27
Fix: SSH bug in windows

SuvarnaNarayanan:
With `SshMachine`, if we were to use windows specific path, it translates into something like `/c/Users/...` which isn't a compatible path for the `scp` call in windows. I have changed the logic that calls the translation function to run only when the underlying platform isn't windows, which fixes the problem.

Also, some basic test cases fix to prevent `ImportError` when trying to get `printenv`

### DIFF
--- a/plumbum/machines/ssh_machine.py
+++ b/plumbum/machines/ssh_machine.py
@@ -129,7 +129,7 @@ class SshMachine(BaseRemoteMachine):
                         Ctrl+C (SIGINT)
     """
 
-    __slots__ = ("_fqhost", "_scp_command", "_ssh_command", "host")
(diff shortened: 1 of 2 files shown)


tests/test_env.py
```diff
@@ -5,7 +5,7 @@
 from plumbum import local
 from plumbum._testtools import skip_on_windows
 
-with contextlib.suppress(ModuleNotFoundError):
+with contextlib.suppress(ModuleNotFoundError, ImportError):
```
henryiii:
```suggestion
with contextlib.suppress(ImportError):
```

`ModuleNotFoundError` is an `ImportError`

Not sure about the other change yet, just noting this quickly.

Issue #530, opened 2020-12-31
Plumbum should have asyncio support


Callek:
I feel that being able to use await/etc syntax with plumbum would be helpful in a lot of areas, especially for remote machine use.  I may be able to work on implementing this in my work time but would love insight on if this would be accepted by the maintainer and any thoughts on how this should be implemented.

Callek:
@henryiii - What do you think?

henryiii:
I've thought about this before, but haven't had the time or enough familiarity with async (seems like I never get past making simple examples and demos, it hasn't quite been enough for ). Ideally, I assume this would be useful either as a base or a guide: https://docs.python.org/3/library/asyncio-subprocess.html.

I have a few long term plans (but not enough time to do them): I'd like to break off at least `plumbum.cli` and `plumbum.colorlib` into sub packages that are available separately, I'd like to see if paths can be refactored to use pathlib (with remote paths being a subclass), and somewhere in this, after 1.7, we could drop Python 2 support as long as `python_requires` is handled correctly (I'd be okay with 3.6+). I could see async, pathlib, typing, and refactoring of commands (I think we recreate things that might be able to use high level constructs in modern Python) as a nice 2.0 project.

Callek:
If I'm reading this right you are considering the async support replacing the sync support, for a 2.0 after the 1.7 stuff is done?

henryiii:
It would be in addition to it, not a replacement - in fact, it seems possible to detect if something is being called in await, so it might be possible to be pretty elegant in allowing both. I was referring more to the fact we could possibly keep the backends from diverging too badly by updating the sync backend to look like the async backend (assuming we can use the higher level abstractions now).

henryiii:
If you want to draft a basic proposal and/or proof of concept PR, I'd be happy to review.

Callek:
Skimming the code so far, it looks like some of the issues with implementing this is going to be almost a rewrite.

For example, we should ideally allow `await Foo()...` for any call that would utilize a network request, that said even `ShellSession` does a connection and a run from `__init__` and its used directly from RemoteMachine's init.

Do you have any suggestions on what the base areas of code would be that I should work on updating or changing here?

Callek:
@henryiii 

So my basic proposal would be for async support to have __init__ without side affects as part of the design. This would be a breaking change even for sync users...

Though we can move the side affects to many common uses, e.g. contextmanager could do the connection __init__ does.

[...]

PR #768, opened 2026-01-12
feat: support async mode

yajo:
Fix https://github.com/tomerfiliba/plumbum/issues/530. Now we can use Python's `async`/`await` syntax.

Some implementation details:

## Why Delegation?
The implementation uses **delegation over inheritance** to maximize code reuse.

Sync and async methods are fundamentally incompatible in Python. You cannot override a sync method with an async one in the same class hierarchy. By using separate wrapper classes that delegate to sync commands, we:

1. Reuse all battle-tested logic from sync commands
2. Maintain a single source of truth for command formulation
3. Ensure changes to sync code automatically benefit async code
4. Keep the codebase maintainable and consistent

## Why No BG, FG, NOHUP in Async?

These sync modifiers are not needed in async mode:

- **BG (Background)**: Async commands are already non-blocking by nature
- **FG (Foreground)**: Async I/O handles stdin/stdout/stderr differently
- **NOHUP**: Async processes are already detached from the parent

For these use cases, use sync commands with the appropriate modifiers.

### Basic Command
```python
result = await async_local["echo"]("hello")
print(result)  # "hello\n"
```

### With Full Control
```python
result = await async_local["ls"].run(
    args=["-la"],
    timeout=5.0,
    cwd="/tmp",
    env={"CUSTOM_VAR": "value"}
)
print(result.stdout)
```

### Pipelines
```python
pipeline = async_local["ls"] | async_local["grep"]["py"]
result = await pipeline.run()
```

### Error Handling
```python
try:
    await async_local["false"].run()
except ProcessExecutionError as e:
    print(f"Command failed with code {e.retcode}")
```

### Concurrent Execution
```python
results = await asyncio.gather(
    async_local["echo"]("task1"),
    async_local["echo"]("task2"),
    async_local["echo"]("task3"),
)
```

### Remote Commands
```python
async with AsyncSshMachine("hostname") as rem:
    # Simple command
    result = await rem["ls"]("-la")
    
    # Pipeline
    result = await (rem["ls"] | rem["grep"]["py"])()
    
    # Concurrent execution
    results = await asyncio.gather(
        rem["echo"]("task1"),
        rem["echo"]("task2"),
    )
```

### Execution Modifiers
```python
from plumbum.commands.async_ import AsyncTF, AsyncRETCODE, AsyncTEE

# Check if file exists
exists = await (async_local["test"]["-f", "file.txt"] & AsyncTF)

# Get exit code
code = await (async_local["ls"]["/nonexistent"] & AsyncRETCODE)

# Display and capture output
retcode, stdout, stderr = await (async_local["echo"]["hello"] & AsyncTEE)
```

---

@moduon MT-6769

### DIFF
--- a/plumbum/commands/async_.py
+++ b/plumbum/commands/async_.py
@@ -57,7 +57,7 @@ import asyncio
 import sys
 from typing import TYPE_CHECKING, Any
 
-from plumbum.commands.processes import ProcessExecutionError
+from plumbum.commands.processes import ProcessExecutionError, ProcessTimedOut
 from plumbum.machines.local import local
 
 if sys.version_info >= (3, 11):
@@ -194,7 +194,10 @@ class AsyncCommandMixin:
             )
             return retcode_val, stdout, stderr
 
-        retcode_val, stdout, stderr = await loop.run_in_executor(None, _run_sync)
+        try:
+            retcode_val, stdout, stderr = await loop.run_in_executor(None, _run_sync)
+        except ProcessTimedOut as e:
+            raise asyncio.TimeoutError() from e
 
         return AsyncResult(retcode_val, stdout, stderr)
 
--- a/plumbum/commands/base.py
+++ b/plumbum/commands/base.py
@@ -11,7 +11,11 @@ from types import MethodType
 from typing import ClassVar
 
 import plumbum.commands.modifiers
-from plumbum.commands.processes import iter_lines, run_proc
+from plumbum.commands.processes import (
+    ProcessTimedOut,
+    iter_lines,
+    run_proc,
+)
 
 if typing.TYPE_CHECKING:
     from collections.abc import Container, Generator, Sequence
@@ -459,6 +463,16 @@ class Pipeline(BaseCommand):
             stdout: str,
             stderr: str,
         ) -> None:
+            # Check if any process in the pipeline has timed out
+            # If so, raise ProcessTimedOut instead of ProcessExecutionError
+            if getattr(proc.srcproc, "_timed_out", False) or getattr(
+                proc, "_timed_out", False
+            ):
+                raise ProcessTimedOut(
+                    f"Process did not terminate within {timeout} seconds",
+                    getattr(proc, "argv", None),
+                )
+
             # TODO: right now it's impossible to specify different expected
             # return codes for different stages of the pipeline
             try:


yajo:
Yes, the point is that we can't have both sync and async versions of the same function. So this way seems a lot more ergonomic, because you just get the commands in async mode and then keep working with them in async mode just like you always did with plumbum.

That's the reason also why I used delegation instead of inheritance.

Sorry for the 3000 lines... 😅 most are tests and docs.

docs/async_support.rst
```diff
@@ -0,0 +1,457 @@
+Async Support
+=============
+
+.. versionadded:: 2.0
```
yajo: BTW I assumed that you maybe wanted to push a new major version, but in reality this change is completely backwards-compatible, so maybe this should be:


```suggestion
.. versionadded:: 1.11
```

plumbum/machines/local.py
```diff
@@ -490,3 +490,80 @@ def as_root(self) -> AbstractContextManager[None]:
...
+    def cwd(self) -> LocalPath:
+        """Current working directory."""
+        return local.cwd  # type: ignore[no-any-return]
+
+    @property
+    def env(self) -> LocalEnv:
+        """Environment variables."""
+        return local.env
+
+    def path(self, *parts: str) -> Any:
```
henryiii: This could be a `@staticmethod`.

docs/async_support.rst
henryiii: Might be better to assume 1.11 for now, and can change later. I'd be pretty tempted to try to rework path handling for 2.0 to get rid of string inheritance and make it fully pathlib compatible.

plumbum/machines/remote.py
```diff
@@ -506,3 +507,80 @@ def expanduser(self, expr: str) -> str:
...
+
+    Example::
+
+        from plumbum.machines.ssh_machine import AsyncSshMachine
+
+        async with AsyncSshMachine("host") as rem:
+            ls = rem["ls"]
+            result = await ls("-la")
+
+    .. versionadded:: 2.0
```
henryiii:
```suggestion
    .. versionadded:: 1.11
```

plumbum/machines/ssh_machine.py
```diff
@@ -468,3 +467,112 @@ def session(self, isatty: bool = False, new_session: bool = False) -> ShellSessi
...
+        async with AsyncSshMachine("hostname") as rem:
+            result = await rem["ls"]("-la")
+
+            # Concurrent execution
+            results = await asyncio.gather(
+                rem["echo"]("task1"),
+                rem["echo"]("task2"),
+            )
+
+    .. versionadded:: 2.0
```
henryiii:
```suggestion
    .. versionadded:: 1.11
```

plumbum/machines/ssh_machine.py
```diff
@@ -468,3 +467,112 @@ def session(self, isatty: bool = False, new_session: bool = False) -> ShellSessi
             isatty,
             self.connect_timeout,
         )
+
+
+class AsyncSshMachine:
```
henryiii, line 472: I think this could be in a new file `async_ssh_machine.py`), no need to change it here, but could be a follow up. (Unless there's some reason it needs to be in this file, which I don't see.)

yajo: Well, I'm keeping AsyncLocalMachine in `local.py`, so it seems natural to do the same for this one.

I'm happy if you merge it and you can organize code later according to your likings.

henryiii:
2.0 is out!

PR #800, opened 2026-04-14
feat: support async pipes

### LINKED ISSUE #795: [Feature Request] support `popen` for Async Pipeline

Sorry for the poor english :<

## System version

`Python version`: 3.12.12

## Reproduce code & Result

```python
import asyncio
import sys

cmd2 = """
import sys
for line in sys.stdin:
    print(line.strip().upper())
"""

def sync_test():

    command = echo | upper

    proc = command.popen()
        print(repr(i))

async def async_test():

    command = echo | upper

    proc = await command.popen()
        print(repr(i))

async def main():
    print("\n\nSync test:")
    sync_test()
    print("\n\nAsync test:")
    await async_test()

if __name__ == "__main__":
    asyncio.run(main())
```

Result:
```bash


Sync test:
b'TEST PIPE1\r\n'
b'TEST PIPE2\r\n'


Async test:

```

Expect result:
```bash


Sync test:
b'TEST PIPE1\r\n'
b'TEST PIPE2\r\n'


Async test:
b'TEST PIPE1\r\n'
b'TEST PIPE2\r\n'

```

henryiii:
Fix #795.

:robot: Assisted-by: Copilot:claude-sonnet-4.6


yajo (commented):
## Automated Review Summary

This PR adds `AsyncPipeline` and `AsyncPipelineProcess` to enable true async OS-pipe pipelines (fixing #795). The core `popen()` implementation is generally sound. However, there are **4 bugs** that should be addressed before merging and several **test gaps** worth considering.

### Bugs to Fix

1. **`communicate()` can leak tasks on write/drain failure** — `stdin` is not closed in a `finally` block, so if `write()` or `drain()` raises, stdout/stderr tasks hang.
2. **`AsyncPipeline` doesn't override `run()`/`__call__()`** — calls to `await (cmd1 | cmd2)()` or `await pipeline.run()` fall back to the sync `Pipeline` run in a thread pool via `AsyncCommandMixin.run()`, completely bypassing the new OS-pipe implementation. This is the main entry point for most users.
3. **Signal tests are flaky** — they use `asyncio.sleep(0.1)` and check `proc._procs[0].returncode` directly (private attr). Will fail on slow CI.
4. **Dead code in tests** — `echo_cmd_template` is defined but never used; one line in `test_async_pipeline_getitem_binds_to_right_command` creates a command and discards it immediately.

### Test Gaps (blocking usability)

5. **No test for `run()`/`__call__()` on pipeline** — the primary usage path for most users. Without this, the bug in #2 is silent.
6. **No test for pipelines with 4+ stages** — only 3-stage pipelines are exercised.
7. **No test for interleaved stdin/stdout interaction** — the real-world use case from #795 (progress bars) involves writing to stdin while reading stdout concurrently.
8. **No test for concurrent pipelines** — the stated goal ("run multiple processes lightweightly") is not tested. With `os.pipe()` buffers, concurrent pipelines can deadlock if stdout isn't consumed promptly.

### Minor / Design Notes

9. **`popen()` ignores `**kwargs`** — users cannot pass `stdin=...`, `stderr=...`, etc.
10. `__getitem__` return type annotation changed from `AsyncCommand` to `AsyncCommandMixin` but always returns `AsyncCommand` — annotation over-promises.
11. `AsyncPipelineProcess.__getattr__` delegates to the last process — should be documented.
12. `kill()`/`terminate()` use `__import__("signal")` inline instead of a module-level import.

---

*This is an automated review performed by minimax-m2.7 (via opencode).*

yajo:
## Bug 1: `communicate()` can leak tasks if `stdin.write` fails

**File:** `plumbum/commands/async_.py`, lines ~157-162

If `self.stdin.write(input)` or `await self.stdin.drain()` raises (e.g., `BrokenPipeError` because the first process exited early), `self.stdin` is never closed. The `stdout_task` and `stderr_task` created earlier will hang indefinitely waiting for EOF.

**Suggested fix:**
```python
if input is not None:
    try:
        self.stdin.write(input)
        await self.stdin.drain()
    finally:
        self.stdin.close()
```

---

*Automated review by minimax-m2.7 (opencode)*

yajo:
## Minor: `kill()` and `terminate()` use `__import__("signal")` inline

Instead of calling `__import__("signal")` on every invocation, import `signal` at module level once and reuse `signal.SIGTERM`/`signal.SIGKILL`.

---

*Automated review by minimax-m2.7 (opencode)*

yajo:
## Bug 3: Signal tests are flaky and access private attributes

**File:** `tests/test_issue_795.py`, lines ~324, 334, 344

`test_async_pipeline_terminate`, `test_async_pipeline_kill`, and `test_async_pipeline_send_signal` all use `await asyncio.sleep(0.1)` then check `proc._procs[0].returncode is not None`. Problems:

1. On slow CI runners, 100ms may not be enough for the OS to update `returncode`.
2. Accessing `_procs` is a test smell — it's a private attribute.
3. No assertion on what the returncode actually is (could be -15 for SIGTERM or -9 for SIGKILL — they should be checked).

**Suggested fix:** Use `asyncio.wait_for(proc._procs[0].wait(), timeout=5.0)` instead of the fixed sleep, and assert on the actual signal number received.

---

*Automated review by minimax-m2.7 (opencode)*

yajo:
## Dead code in tests

**File:** `tests/test_issue_795.py`

1. Line 9: `echo_cmd_template = "print({!r}\ncat 'test pipe2')"` is defined but never used. Should be removed.

2. `test_async_pipeline_getitem_binds_to_right_command` (line ~72): the first line `async_local[sys.executable]["-c", "import sys; print(sys.argv[1].lower())"]` creates an `AsyncCommand` and immediately discards it. Should be removed.

---

*Automated review by minimax-m2.7 (opencode)*

yajo:
## Missing test: `run()` and `__call__()` on `AsyncPipeline`

The primary entry point for most users will be `await (cmd1 | cmd2)()` or `await pipeline.run()`, not `await pipeline.popen()` directly. However, there is no test covering these paths. More critically, `AsyncPipeline` doesn't override `run()`/`__call__()`, so these currently fall back to the sync `Pipeline.run()` via `loop.run_in_executor()` — completely bypassing the OS-pipe implementation. This should be tested before merging.

---

*Automated review by minimax-m2.7 (opencode)*

yajo:
## Missing tests: 4+ stage pipelines

Only 3-stage pipelines are exercised in the test suite (`cat | upper | cat2`). There is no test for 4, 5, or N stages. While the `_flatten_pipeline()` recursion should handle this, it's not verified. Consider adding at least one test for a 4-stage pipeline to ensure correctness.

---

*Automated review by minimax-m2.7 (opencode)*

yajo:
## Missing test: interleaved stdin/stdout (real-world use case from #795)

The original issue author wants `popen()` for real-time progress bars, which requires writing to stdin while reading stdout concurrently — not a single `communicate()` call. There's no test exercising this interleaved pattern:

```python
proc = await pipeline.popen()
proc.stdin.write(b"input1\n")
await proc.stdin.drain()
line1 = await proc.stdout.readline()
proc.stdin.write(b"input2\n")
await proc.stdin.drain()
line2 = await proc.stdout.readline()
```

This pattern is also where `os.pipe()` buffer deadlocks can occur if stdout isn't consumed while stdin is being written to.

---

*Automated review by minimax-m2.7 (opencode)*

yajo:
## Missing test: concurrent pipelines

The stated goal of the original issue (#795) is to "run multiple processes more lightweightly without the need to manage multiple threads." However, there's no test exercising concurrent pipelines with `asyncio.gather()`. This is important because `os.pipe()` has a limited buffer (~64KB on Linux); if multiple pipelines write to stdin while not consuming stdout promptly, they can deadlock.

---

*Automated review by minimax-m2.7 (opencode)*

yajo:
## Design note: `popen()` ignores `**kwargs`

`AsyncCommandMixin.popen()` accepts `**kwargs` but `AsyncPipeline.popen()` does not. This means users cannot pass `stdin=...`, `stderr=...`, `close_fds=False`, etc. when using async pipelines. Consider accepting and forwarding `**kwargs` in a follow-up, or at minimum document this limitation.

---

*Automated review by minimax-m2.7 (opencode)*

PR #802, opened 2026-04-20
feat: support async pipe

### LINKED ISSUE #795: [Feature Request] support `popen` for Async Pipeline

henryiii:
Trying Kimi-K2.6, which came out earlier today. It thought for nearly an hour, then wrote this solution.

I've been using the same tests, just asking for a solution to the issue.

Fix #795.

:robot: Assisted-by: OpenCode:Kimi-K2.6

### DIFF
--- a/plumbum/commands/async_.py
+++ b/plumbum/commands/async_.py
@@ -232,6 +232,21 @@ class AsyncCommandMixin:
         base_cwd = getattr(self._base_cmd, "cwd", None)
         working_dir = cwd or base_cwd or str(local.cwd)
 
+        # If the formulated command contains a "|" shell operator, it represents
+        # a pipeline. We must use a shell to properly connect the processes.
+        is_pipeline = "|" in argv
+
+        if is_pipeline:
+            shell_cmd = " ".join(argv)
+            return await asyncio.create_subprocess_shell(
+                shell_cmd,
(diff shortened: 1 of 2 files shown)


henryiii:
Interesting - it thought for ~40 mins and then came up with this, which is really short, but "wrong", in that it really just converts the command to a string then runs it in the shell (which works on non-windows systems).

PR #804, opened 2026-06-01
feat(async): support popen on async pipelines (#795)

### LINKED ISSUE #795: [Feature Request] support `popen` for Async Pipeline

henryiii:
Using Claude Code, Opus 4.8 + Copilot review, 7 iteration battle (8th iteration passed review). Burned about 25% of my monthly quota on the reviews with the new copilot pricing!

* Closes #795.
* Closes #796. First attempt with Copilot GPT-5-mini. (0x) Very basic and long.
* Closes #800. Second attempt with Copilot Sonnet 4.6. (1x) More elegant but long.
* Closes #802. Third attempt with OpenCode Kimi K2.6. It cheated.

This is pretty close to the sonnet one now, I think, but I trust this one more because I saw why everything is needed. :)

:robot: _AI text follows_ :robot:

## Problem

Calling `popen()` on an **async** pipeline didn't actually pipe. `AsyncCommandMixin.popen` formulated the pipeline into a single argv with a literal `"|"` argument and exec'd it as one process, so the downstream command never received the upstream's output and `await proc.stdout.readline()` returned nothing.

```python
echo = async_local[sys.executable]["-c", "print('test pipe1\ntest pipe2')"]
upper = async_local[sys.executable]["-c", "...upper..."]
proc = await (echo | upper).popen()
while i := await proc.stdout.readline():  # never yielded any lines
    ...
```

## Fix

Mirror the synchronous `Pipeline.popen`:

- The public `popen` now delegates to an internal `_popen` that exposes `stdin`/`stdout`/`stderr` so pipeline stages can be chained.
- When the wrapped command is a `Pipeline`, the two stages are connected with an `os.pipe()` — the source's stdout writes the write end, the destination reads the read end. (asyncio can't connect a `StreamReader` to another process's stdin, so an OS-level pipe is the clean equivalent of what a shell does.)
- Recurses through the wrapped sync commands, so nested pipelines (`a | b | c`) work automatically.
- `dstproc.wait()` is wrapped to also reap the upstream process. As in a shell, the pipeline's return code is that of its last stage.

The public API is unchanged (`popen(args, cwd, env)` still uses pipes for all three streams), and the single-command branch reuses the existing env/cwd resolution.

## Tests

Added `TestAsyncPipeline.test_pipeline_popen` (the issue #795 reproducer) which streams `readline()` over a piped pair of processes and asserts both children are reaped with `returncode == 0`. Full async suite passes; `prek -a` (ruff + mypy strict + codespell) is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Assisted-by: ClaudeCode:claude-opus-4.8

### DIFF
--- a/docs/async_support.rst
+++ b/docs/async_support.rst
@@ -338,7 +338,27 @@ AsyncCommand
 
       Create an async subprocess without waiting for completion.
 
-      :return: asyncio.subprocess.Process instance
+      :return: an :class:`asyncio.subprocess.Process` for a plain command, or an
+         :class:`AsyncPipelineProcess` proxy for a pipeline.
+
+AsyncPipelineProcess
+~~~~~~~~~~~~~~~~~~~~~
+
+.. class:: AsyncPipelineProcess
+
+   Returned by :meth:`AsyncCommand.popen` when called on a pipeline. It behaves
+   like an :class:`asyncio.subprocess.Process`: ``stdout``/``stderr`` come from
+   the last stage and ``stdin`` writes to the first stage. ``wait()`` and
+   ``communicate()`` reap every stage and report a combined return code (the
+   last stage's, or an earlier stage's if the last one succeeded), and
+   ``kill()``/``terminate()``/``send_signal()`` are sent to every stage.
+
+   .. attribute:: srcproc
+
+      The upstream process (itself an :class:`AsyncPipelineProcess` for a nested
+      pipeline).
+
+   .. versionadded:: 1.11
 
 AsyncResult
 ~~~~~~~~~~~
--- a/plumbum/commands/async_.py
+++ b/plumbum/commands/async_.py
@@ -54,6 +54,7 @@ Example Usage
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import os
 import sys
 from typing import TYPE_CHECKING, Any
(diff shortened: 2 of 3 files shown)


henryiii:
> generated no new comments.

😓 Finally! This would have been so much faster with Copilot's rubber duck mode...

PR #813, opened 2026-06-02
feat: add @cli.configs decorator for argument config files (fixes #714)

### LINKED ISSUE #714: Feature request: Argument config files
It would be nice to be able to specify config files containing CLI args, and when a CLI app is ran the args in those files would be added on (though with duplicate args being overwritten). This also helps more closely Integrate config files and the arg parser.

An example of what I mean:

```py
# script.py

from plumbum import cli, local

class MyApp(cli.Application):
    myarg = cli.SwitchAttr(["--myarg"], str, mandatory=True)

    @cli.configs(local.path("~/.config/myapp/config.cfg"))
    def main(self):
        print("myarg: ", self.myarg)

MyApp.run()
```

```ini
# ~/.config/myapp/config.cfg
--myarg=foobar
```

```sh
$ python script.py
myarg: foobar
$ python script.py --help
Usage:
    script.py [SWITCHES] args...

Argument config files:
    ~/.config/myapp/config.cfg
```

mshzy:
feat: add @cli.configs decorator for config file support

Add a `configs` classmethod decorator that registers config files
containing CLI arguments. Config file args are processed before CLI
args, so CLI arguments take precedence.

Features:
- `@cli.configs(path1, path2)` decorator on Application subclasses
- Config files: one argument per line, `#` comments supported
- Help output lists registered config files
- Missing config files are silently skipped

Fixes #714

### DIFF
--- a/plumbum/cli/__init__.py
+++ b/plumbum/cli/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from .application import Application
+from .application import configs  # noqa: F401
 from .config import Config, ConfigINI
 from .switches import (
     CSV,
@@ -22,6 +23,7 @@ from .switches import (
 __all__ = (
     "CSV",
     "Application",
+    "configs",
     "Config",
     "ConfigINI",
     "CountOf",
--- a/plumbum/cli/application.py
+++ b/plumbum/cli/application.py
@@ -190,6 +190,7 @@ class Application:
         "see '{parent} {sub} --help' for more info"
     )
     ALLOW_ABBREV: bool = False
+    CONFIG_FILES: ClassVar[tuple[str, ...]] = ()
 
     parent: Self | None = None
     nested_command: tuple[type[Application], list[str]] | None = None
@@ -291,6 +292,49 @@ class Application:
         cls._unbound_switches += tuple(
             name.lstrip("-") for name in switch_names if name
         )
+
+    @classmethod
+    def configs(cls, *paths: str) -> Callable[[type[Application]], type[Application]]:
+        """Decorator that registers config files for this application.
+
+        Config files contain CLI arguments (one per line or space-separated).
+        Lines starting with ``#`` are treated as comments.
+        Arguments from config files are processed first; CLI arguments override them.
+
+        Usage::
+
+            @MyApp.configs("~/.config/myapp/config.cfg", "/etc/myapp.conf")
+            class MyApp(cli.Application):
+                pass"""
+
+        def wrapper(app_cls: type[Application]) -> type[Application]:
+            app_cls.CONFIG_FILES = app_cls.CONFIG_FILES + paths
+            return app_cls
+
+        return wrapper
+
+    @staticmethod
+    def _read_config_files(config_files: tuple[str, ...]) -> list[str]:
+        """Read config files and return CLI-style argument list."""
+        import shlex
+        args: list[str] = []
+        for path in config_files:
+            expanded = os.path.expanduser(path)
+            if not os.path.isfile(expanded):
+                continue
+            try:
+                with open(expanded, encoding="utf-8") as f:
+                    for line in f:
+                        line = line.strip()
+                        if not line or line.startswith("#"):
+                            continue
+                        try:
+                            args.extend(shlex.split(line))
+                        except ValueError:
+                            args.append(line)
+            except OSError:
+                pass
+        return args
 
     @typing.overload
     @classmethod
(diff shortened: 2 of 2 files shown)


henryiii:
Interesting idea, but several problems (PR review via Claude below if you don't mind AI generated text). I'm not sure we should do just simple args - supporting TOML would probably be better (simple args could just be expanded on the command line, like `cat args.txt | xargs ./app`).

I can throw AI at this after 2.0 is out if you are interested in TOML support, probably mirroring CLI11.

<details><summary>:robot: AI review :robot:</summary>
<p>

# Code Review: PR #813 — `@cli.configs` decorator for argument config files

## Overview
Adds a way to load CLI arguments from config files. Implements `Application.configs(*paths)` as a classmethod decorator that appends to a new `CONFIG_FILES` class attribute; `run()` reads those files and prepends their tokens to `argv` so CLI args "take precedence." Help output lists the registered files. Fixes #714.

The idea is sound and useful, but **the PR is not functional as written** — there are two blocking bugs, and the central design premise (CLI overrides config) doesn't actually hold.

## 🔴 Blocking issues

**1. The package fails to import at all.**
`plumbum/cli/__init__.py` does `from .application import configs`, but `configs` is only defined as a **classmethod on `Application`**, not a module-level name. I confirmed this crashes on import:
```
ImportError: cannot import name 'configs' from 'plumbum.cli.application'
```
This breaks `import plumbum.cli` entirely — every CLI user, not just config-file users. The `__all__` entry has the same problem. Either export `Application.configs` differently or add a real module-level `configs` function.

**2. The documented usage is invalid Python.** The docstring shows:
```python
@MyApp.configs("~/.config/myapp/config.cfg", "/etc/myapp.conf")
class MyApp(cli.Application):
    pass
```
`MyApp` isn't defined yet at decoration time → `NameError`. You can't decorate a class with its own classmethod. This also diverges from the API requested in #714, which is a module-level `@cli.configs(...)` decorator on `main`. The intended call shape needs to be settled and made actually runnable.

**3. "CLI arguments override config args" is false for normal switches.** This is the core feature promise, and it doesn't work. Prepending tokens to `argv` means a switch supplied by *both* the config file and the command line is seen twice by the parser. For any non-`list` switch, `_parse_args` raises rather than overriding (application.py:497-503):
```python
raise SwitchError(T_("Switch {0} already given").format(swname))
```
So `myapp --myarg=bar` with `--myarg=foo` in the config file **errors out** instead of letting the CLI win. The prepend trick only works for switches the user *omits* on the CLI (config-as-default). Real override semantics would require deduplicating switches before parsing, or threading config values in at a lower level. At minimum the description/docs must stop claiming override.

## 🟡 Other concerns

- **Zero test coverage.** Only the 2 source files changed; nothing in `tests/`. A feature touching `run()` and arg parsing needs tests — file parsing, `#` comments, missing files, `~` expansion, precedence, and the duplicate-switch case above. This is also how bugs #1–#3 would have been caught.
- **No docs.** Nothing under `docs/`. New public API should be documented.
- **i18n not regenerated.** You added `T_("Argument config files:")` but didn't run `./translations.py` to update `.pot`/`.po`/`.mo` (per CLAUDE.md). The string won't be translatable.
- **Silent error swallowing.** `except OSError: pass` and the missing-file skip hide real problems (permission denied, unreadable file) identically to "no config." Consider warning on read failures vs. silently skipping only genuinely-absent files.
- **`import shlex` inside the function.** Move to the top of the module with the other imports to match project style (`os`, `sys`, etc. are all top-level).
- **Help output bypasses the formatter.** It uses bare `print()` rather than the existing help-rendering machinery in `help()`, so it won't respect width/wrapping or color styling used elsewhere.
- **`line = line.strip()` reassigns the loop var** — harmless but ruff may flag; minor.

## Suggestions
1. Fix the import/export so the package loads, and pin down a single, valid decorator API (likely a module-level `configs` matching #714).
2. Decide on real precedence semantics and implement dedup, or downgrade the claim to "config provides defaults; explicitly-passed switches still conflict."
3. Add tests covering parsing, comments, `~`, missing files, and precedence.
4. Add docs + regenerate translations.

The feature is worth having, but as-is it doesn't import, the documented usage doesn't run, and the headline behavior doesn't work for the common case. I'd request changes before merge.

</p>
</details>

mshzy:
Thanks for the thorough review @henryiii! The AI review correctly identified several blocking issues — the import failure, invalid decorator usage, and the duplicate-switch problem are all real bugs that would need significant rework.

You're right that TOML support would be a better direction for this feature. I'll close this PR for now and look forward to the TOML-based config support post-2.0. Thanks for taking the time to review!

henryiii:
I _will_ need to be prodded and poked regularly, or I'll forget to do it! :)

PR #817, opened 2026-06-08
fix: make SwitchInfo frozen dataclass to fix unhashable type error with excludes/requires

### LINKED ISSUE #816: TypeError: unhashable type: 'SwitchInfo' when using excludes= or requires= on a switch
## Bug

Since plumbum 2.0.0, any application that uses `excludes=` or `requires=` on a `cli.Flag` or `cli.SwitchAttr` crashes at runtime with:

```
TypeError: unhashable type: 'SwitchInfo'
```

## Root cause

Commit `975d8e6` converted `SwitchInfo` from a named tuple to a `@dataclasses.dataclass`. Python dataclasses generate `__eq__` by default, which sets `__hash__ = None`, making instances unhashable. However, `application.py` builds a `set` of `SwitchInfo` objects when validating switch combinations:

```python
exclusions[swinfo.func] = {
    self._switches_by_name[exc] for exc in swinfo.excludes
}
```

## Reproducer

```python
from plumbum import cli

class MyApp(cli.Application):
    flag_a = cli.Flag("--flag-a")
    flag_b = cli.Flag("--flag-b", excludes=["--flag-a"])

    def main(self):
        pass

MyApp.run()
```

```
TypeError: unhashable type: 'SwitchInfo'
```

## Fix

Change `SwitchInfo` to a frozen dataclass, which generates a correct value-based `__hash__` and enforces immutability:

```python
@dataclasses.dataclass(frozen=True)
class SwitchInfo:
```

zhan9san:
## Summary

- `SwitchInfo` was converted to a plain `@dataclasses.dataclass` in `975d8e6`, which sets `__hash__ = None` (because `__eq__` is auto-generated), making instances unhashable
- `application.py` builds a `set` of `SwitchInfo` objects when validating `excludes=`/`requires=` switch combinations, causing `TypeError: unhashable type: 'SwitchInfo'`
- Fix: use `@dataclasses.dataclass(frozen=True)` which generates a correct value-based `__hash__` and enforces the natural immutability invariant of switch metadata

Closes #816

## Test plan

- [ ] `test_excludes_no_crash` — regression test: defining `excludes=` no longer crashes with `unhashable type`
- [ ] `test_excludes_raises_on_conflict` — using two mutually exclusive flags raises `SwitchCombinationError`
- [ ] `test_requires_raises_when_missing` — using a flag without its required companion raises `SwitchCombinationError`
- [ ] `test_requires_ok_when_present` — valid combination passes cleanly

### DIFF
--- a/plumbum/cli/switches.py
+++ b/plumbum/cli/switches.py
@@ -67,7 +67,7 @@ T = TypeVar("T", default=str)
 # ===================================================================================================
 # The switch decorator
 # ===================================================================================================
-@dataclasses.dataclass
+@dataclasses.dataclass(frozen=True)
 class SwitchInfo:
     # Python 3.10+ can use slots=True
     __slots__ = (
--- a/tests/test_3_cli.py
+++ b/tests/test_3_cli.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
+import pytest
+
 from plumbum import cli
+from plumbum.cli.application import SwitchCombinationError
 
 
 class Main3Validator(cli.Application):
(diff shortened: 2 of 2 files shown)


zhan9san:
@henryiii 

It fixes the bug in [artifactory-cleanup](https://github.com/devopshq/artifactory-cleanup/issues/181)

Would you plz help review this PR?
Thanks

henryiii:
I believe #818 is the correct fix, I believe it exactly restores the old behavior.

PR #818, opened 2026-06-08
Make SwitchInfo hashable again so requires=/excludes= work (#816)

### LINKED ISSUE #816: TypeError: unhashable type: 'SwitchInfo' when using excludes= or requires= on a switch

gaoflow:
Fixes #816.

## Problem

Since 2.0.0, any app using `requires=` or `excludes=` on a `Flag`/`SwitchAttr` crashes at parse time with:

```
TypeError: unhashable type: 'SwitchInfo'
```

`SwitchInfo` was changed from a `namedtuple` into a plain `@dataclass`. A value-comparing dataclass (the default, `eq=True`) sets `__hash__ = None`, but `Application` collects `SwitchInfo` objects into sets when checking `requires`/`excludes`:

```python
exclusions[swinfo.func] = {self._switches_by_name[exc] for exc in swinfo.excludes}
```

## Fix

`@dataclasses.dataclass(eq=False)` restores the default identity-based `__hash__` (and `__eq__`).

I used identity rather than the `frozen=True` suggested in the issue, because `SwitchInfo` has mutable `list` fields (`names`, `requires`, `excludes`): a value-based hash isn't possible, so `frozen=True` just trades the error for `TypeError: unhashable type: 'list'` the moment the hash is used. Identity is also the correct semantics here — these sets only deduplicate and iterate the canonical switch objects (the actual set algebra runs on `.func`), and the single `==` on `SwitchInfo` (help rendering) compares those same canonical instances, so identity equality is equivalent. I checked the rest of the namedtuple→dataclass change for other fallout (tuple unpacking/indexing) and found none — only the hashing broke.

## Tests

`requires=`/`excludes=` had no test coverage, which is how this slipped through. Added `TestSwitchCombinations` covering both (runs when satisfied, errors when violated) plus a direct hashability check. All five fail on `master` and pass with the fix.

Thanks to @zhan9san for the clear report and root-cause analysis.

### DIFF
--- a/plumbum/cli/switches.py
+++ b/plumbum/cli/switches.py
@@ -67,7 +67,13 @@ T = TypeVar("T", default=str)
 # ===================================================================================================
 # The switch decorator
 # ===================================================================================================
-@dataclasses.dataclass
+# ``eq=False`` keeps the default identity-based ``__hash__`` (and ``__eq__``).
+# The dataclass has mutable ``list`` fields, so it cannot be value-hashed, yet
+# ``Application`` puts ``SwitchInfo`` objects into sets when validating
+# ``requires``/``excludes``. Identity semantics are correct here: each switch is
+# a unique descriptor. (A value-based ``__eq__`` sets ``__hash__ = None``, which
+# is what broke ``requires=``/``excludes=`` when this became a dataclass.)
+@dataclasses.dataclass(eq=False)
 class SwitchInfo:
     # Python 3.10+ can use slots=True
     __slots__ = (
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from plumbum import cli, local
+from plumbum.cli.switches import SwitchInfo
 from plumbum.cli.terminal import get_terminal_size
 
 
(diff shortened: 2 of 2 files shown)


henryiii:
Close #817, this is better. Claude agrees with me too. :)

Thanks for the quick fix!

Issue #820, opened 2026-06-13
Plumbum review by Claude Fable


henryiii:
:robot: _AI text below_ :robot:

All spot-checks confirm the agents' findings. Here's the consolidated review.

# Plumbum code review

Six parallel reviewers covered the whole package (~14.4k lines); I spot-checked the highest-severity claims against the source and they hold up. Many findings were verified by the reviewers with actual reproductions. The short version: the library is in good shape overall, but there's a cluster of **shell-quoting holes in the remote machinery**, several **process-lifecycle bugs** (leaks/deadlocks/broken timeouts), and a handful of **reproduced CLI and colorlib bugs**. Modernization debt is light — mostly small dead compat branches.

## High severity

**Shell injection / quoting in remote machines.** Command *arguments* are carefully shquoted, which makes these stand out:
- `machines/remote.py:625-630` — `expanduser` runs `echo {expr}` through the shell escaping only `$`. It's invoked on every path part starting with `~` via `machine.path()`, so hostile or merely unusual names (backticks, `;`, globs, runs of spaces) execute or mangle. 
- `machines/ssh_machine.py:221` and `paramiko_machine.py:362` — `cwd` is inserted unquoted into the remote command line (`["cd", str(cwd), "&&"]`). After `chdir()` to a directory with a space or shell metacharacters, every subsequent command breaks or injects. Env var *names* are also unquoted (values are quoted), here and in `RemoteEnv.__setitem__`/`__delitem__` (`remote.py:65-79`).
- `machines/remote.py:514-522` — non-recursive `_path_glob` escapes only spaces and uses unquoted `echo $fn`, so directory names with other whitespace/metacharacters inject or come back mangled. The recursive `**` branch was already rewritten safely with `find` + regex; the same technique fixes this branch.

**`async_.py:731` — `_AsyncTEE` dies on long lines and leaks the subprocess.** `await stream.readline()` raises `LimitOverrunError` past the 64 KiB default stream limit. The module's own `drain_stream` (line 745) documents and avoids exactly this — but only for upstream stderr, not the main stdout/stderr. On that exception the `except asyncio.TimeoutError` cleanup is skipped, so the child is neither killed nor reaped and sibling reader tasks leak.

[...]

PR #822, opened 2026-06-13
fix: harden subprocess and pipeline lifecycle handling

henryiii:

Part of the review in #820. Subprocess / pipeline lifecycle robustness.

- **`bgrun` (`commands/base.py`)** — wrapped `yield p` in `try/finally: runner()` so the process is always reaped and pipes closed even when the `with` body raises (incl. KeyboardInterrupt). The existing `was_run` guard keeps reaping idempotent.
- **`CommandNotFound.__init__` (`commands/processes.py`)** — `super().__init__(self, …)` put `self` into `.args`, breaking repr/pickle. Now `super().__init__(program, path)`.
- **Timeout heap** — equal deadlines made `heapq` compare `Popen` objects → `TypeError` that killed the singleton timeout thread and dropped all pending timeouts. Added an `itertools.count()` tiebreaker.
- **`Pipeline.formulate` vs `popen`** — call-time args went to the *source* in `popen` but the *dest* in `formulate` (`(a | b)("x")` → `a x | b` locally but `a | b x` when formulated). `formulate` now matches `popen`.
- **Pipeline source stderr** — closes the previously-leaked source `stderr`/`stdin` pipes. The >64 KiB-stderr *deadlock* is **intentionally not** fixed here: a correct fix conflicts with the `iter_lines` path that reads `srcproc.stderr` itself, and needs broader `run_proc`/`Pipeline` changes — left as a follow-up.
- **async `_AsyncTEE` (`commands/async_.py`)** — `readline()` raised `LimitOverrunError` past the 64 KiB stream limit and leaked the subprocess; now reads fixed chunks via an incremental decoder (also fixes multibyte corruption in `drain_stream`), and cleanup runs on any failure/cancellation.
- **`daemons.py` modernizations** — `errno.ESRCH` → `ProcessLookupError`; EPERM treated as still-running; fake `wait` accepts `timeout`; `subprocess.DETACHED_PROCESS`.

Adds `tests/test_subprocess_lifecycle.py`. `prek -a` clean; pytest green (pre-existing sandbox-only failures unrelated).

### DIFF
--- a/plumbum/commands/async_.py
+++ b/plumbum/commands/async_.py
@@ -722,9 +722,11 @@ class _AsyncTEE(AsyncExecutionModifier):
         stderr_lines: list[str] = []
 
         async def read_stream(
-            stream: asyncio.StreamReader | None, output_list: list[str], target: TextIO
+            stream: asyncio.StreamReader | None,
+            target: TextIO,
+            output_list: list[str] | None = None,
         ) -> None:
-            """Read from stream in fixed-size chunks, display, and collect output.
+            """Read a stream in fixed-size chunks, display, and optionally collect.
 
             Reads by chunk rather than ``readline``: ``StreamReader.readline``
             raises ``LimitOverrunError``/``ValueError`` once a line exceeds the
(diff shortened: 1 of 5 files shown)


henryiii:

A cleanup review pass (reuse / simplification / efficiency / altitude) ran over this branch. Most findings were applied; three were skipped on purpose:

- **`p.wait(timeout=1)` instead of the poll loop in bgrun cleanup** — session and paramiko popen objects declare `wait()` with no timeout parameter, so the stdlib form would break remote commands. The poll loop stays, now inside the shared `_terminate_and_reap()` helper.
- **Restructuring the deadline arithmetic in `daemons.py` `wait()`** — the suggested simpler form loosens the timeout raise by up to 0.5s; the current `min(0.5, remaining)` sleep keeps it tight.
- **Replacing the monkey-patched `wait2` with a real pipeline-Popen wrapper class** — the right long-term fix, but too invasive for this branch.

PR #824, opened 2026-06-13
fix(cli): correct several CLI parsing and terminal bugs

henryiii:

Part of the review in #820. Several CLI parsing and terminal bugs.

- **`Set._call_iter` CSV** — the csv branch didn't `return`, so it also validated the unsplit string: the exported `CSV("a,b")` yielded `['a','b','a,b']`. Fixed.
- **`ALLOW_ABBREV`** — an exact switch name that prefixes another (`--foo` vs `--foobar`) was rejected as ambiguous; exact match now wins (argparse-style).
- **`--flag=value` on a no-arg switch** silently leaked `=value` into positionals; now raises a clear `SwitchError`.
- **`helpall`** permanently mutated shared `SwitchInfo.group` objects process-wide; now computes the demoted grouping locally (regression test asserts the shared object is untouched).
- **`requires=`/`excludes=`** with an unknown name raised a bare `KeyError`; now a clear `SwitchError`.
- **`terminal.ask`/`choose`/`prompt`** infinite-looped on EOF (the caught `EOFError` was never raised by `sys.stdin.readline()`); `readline()` now detects EOF.
- **`Image.best_aspect`** inverted the aspect ratio for non-square images (wide rendered tall); math fixed with a numeric regression test.

The case-insensitive `Set` canonical-spelling behavior change moved to its own PR: #835. This PR is now bug-fixes-only.

`Range.choices()` materialization was left as-is (both callers fully materialize anyway — noted as follow-up). Tests added across `test_cli.py`/`test_validate.py`/`test_terminal.py`. `prek -a` clean; pytest green (pre-existing sandbox-only failures unrelated).

### DIFF
--- a/plumbum/cli/application.py
+++ b/plumbum/cli/application.py
@@ -13,6 +13,7 @@ __lazy_modules__ = {
 }
 
 import contextlib
+import errno
 import functools
 import inspect
 import os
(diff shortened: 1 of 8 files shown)


plumbum/cli/application.py
```diff
@@ -394,10 +411,13 @@ def _parse_args(
                 if eqsign >= 0:
```
henryiii:
```suggestion
                has_eq = eqsign >= 0
                if has_eq:
```

plumbum/cli/application.py
```diff
@@ -394,10 +411,13 @@ def _parse_args(
...
                     name = a[2:eqsign]
                     argv.insert(0, a[eqsign:])
+                    has_eq = True
```
henryiii:
```suggestion
```

plumbum/cli/application.py
```diff
@@ -394,10 +411,13 @@ def _parse_args(
...
                 else:
                     name = a[2:]
+                    has_eq = False
```
henryiii:
```suggestion
```

PR #827, opened 2026-06-25
fix: Path.parent raises IndexError on filesystem root

gaoflow:
## Summary

`LocalPath('/').parent` (and any `Path` subclass at the filesystem root) raised an
`IndexError` instead of returning the root itself.

**Root cause:** `Path.parents` builds its tuple from
`range(len(self.parts) - 1, 0, -1)`.  For the root path `'/'`, `parts` is
`('/',)`, so that range is empty → `parents` is `()`.  The `parent` property then
did `self.parents[0]`, which raised `IndexError`.

**Fix:** return `self` when `parents` is empty, exactly matching
`pathlib.Path('/').parent == PosixPath('/')`.

```python
# Before (raises IndexError)
from plumbum import local
local.path('/').parent

# After
local.path('/').parent  # → LocalPath('/')
```

A regression test is added to `TestLocalPath`.

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/plumbum/path/base.py
+++ b/plumbum/path/base.py
@@ -652,8 +652,13 @@ class Path(str, ABC):
 
     @property
     def parent(self) -> Self:
-        """Pathlib like parent of the path."""
-        return self.parents[0]
+        """Pathlib like parent of the path.
+
+        When called on the root path (e.g. ``/``), returns itself, matching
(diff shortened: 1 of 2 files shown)


henryiii:
Thanks!

PR #828, opened 2026-06-29
Update output to match code

sanxiyn:
When `hex colors` was changed to `rgb colors` in the sample code, the sample output wasn't updated.


henryiii:
Thanks!

Issue #275, opened 2016-04-12
returncode is "Unknown" for command output without trailing newline


horada:
There is a problem with parsing returncode in method [SessionPopen.communicate()](https://github.com/tomerfiliba/plumbum/blob/master/plumbum/machines/session.py#L86).

If the executed command returns some output without newline at the end (for example simple `echo -n "HELLO"`), it will not properly parse the returncode and return *"Unknown"*.

See following reproducer script and check the output for both versions of *cmd* variable (3rd and 4th line, the difference is in the `-n` parameter which will prevent printing the trailing newline).

```python
import plumbum

cmd = "echo -en 'FOO\nBAR'"
#cmd = "echo -e 'FOO\nBAR'"

ssh = plumbum.SshMachine(HOST, USER, \
    keyfile=KEYFILE)
session = ssh.session()

proc = session.popen(cmd)
stdout, stderr = proc.communicate()
rcode = proc.returncode

print("stdout: %s (type: %s)" % (stdout, type(stdout)))
print("stderr: %s (type: %s)" % (stderr, type(stderr)))
print("rcode: %s (type: %s)" % (rcode, type(rcode)))

session.close()
ssh.close()
```
Example output **with** trailing newline:
```
stdout: b'FOO\nBAR\n' (type: <class 'bytes'>)
stderr: b'' (type: <class 'bytes'>)
rcode: 0 (type: <class 'int'>)
```
And the problematic output **without** the trailing newline:
```
stdout: b'FOO\n' (type: <class 'bytes'>)
stderr: b'' (type: <class 'bytes'>)
rcode: Unknown (type: <class 'str'>)
```

https://github.com/tomerfiliba/plumbum/blob/master/plumbum/machines/session.py#L131

henryiii:
I don't have a handy way to test SSH at the moment. I think the problem is related to it working in terms of lines; it should be changed to be able to pick up the sentinel value without a newline. Anyone want to give that a go?

PR #830, opened 2026-07-01
fix(session): keep the return code off the last line of output

### LINKED ISSUE #494: session().run() breaks when the command output does not ends with newline
python 2.7
plumbum 1.6.8

**Command which output does not ends with `'\n'` breaks running with session().run().**

The problem occurs at the method `communicate()` of `plumbum.machines.session.SessionPopen` on the line 144:
```
class SessionPopen
   . . .
    def communicate():
        . . .
        self.returncode = int(stdout.pop(-1))
```

Since the command does not returns '\n', and plumbum echoes the returncode after the command output (`full_cmd += "echo $? ; echo '%s'" % (marker, )`), the readline interpretes the returncode as a part of the command output.
<br/><br/>
**A simple snnipet that breaks:**
```
from plumbum import local

session = local.session()
session.run('echo -n abc')
```
Plumbum will generate the following command:
```
echo -n abc ; echo $? ; echo \'--.END648391697.571.--\' ; echo \'--.END648391697.571.--\' 1>&2
```
And it's output will be:
```
abc0
'--.END648391697.571.--'
'--.END648391697.571.--'
```

apoorva-01:
Fixes #494

The session trailer runs `<cmd> ; echo $? ; echo '<marker>'`, so when a command's output has no trailing newline the return-code digits land on that last line. For `printf abc` the shell writes `abc0`, `int("abc0")` throws, and we fall into the `except` that sets `returncode = "Unknown"` and drops the real output. `session().run("printf abc")` currently returns `("Unknown", "")` instead of `(0, "abc")`.

The fix prints the return code on its own line (`printf '\n%s\n' "$?"`) so it can't share a line with the output, then strips that one injected newline back off. `printf` rather than a bare `echo` because a command in between would reset `$?` first; the strip is tty-aware since a pty turns the injected `\n` into `\r\n`. This is the one marker path that `local`/`SshMachine`/`ParamikoMachine` sessions all share, so it covers local and remote alike.

Regression test in `tests/test_local.py::test_session_no_trailing_newline`, fails on `master` and passes with the fix. Full local suite is green plus ruff/mypy/pylint; the optional SSH suite I couldn't run locally, but `test_remote.py::test_session` exercises the remote path in CI.

One call for you: this is the contained version and leaves the identical stderr-marker case alone to keep scope tight. If you'd rather fix both, the more general route is the reader-level one Henry hinted at on #275 (pick up the sentinel without depending on a newline). Happy to go that way instead.

### DIFF
--- a/plumbum/machines/session.py
+++ b/plumbum/machines/session.py
@@ -215,13 +215,14 @@ class SessionPopen(PopenAddons):
             self.returncode = "Unknown"  # type: ignore[assignment]
         self._done = True
         stdout_bytes = b"".join(stdout)
-        # the command trailer prints the return code on its own line, so it always
-        # appends a newline to the real output; strip it to return the output verbatim
-        # (e.g. `printf abc` yields b"abc", not b"abc\n"). Under a tty the injected
-        # newline is translated to CRLF, so strip that instead of eating a real "\r".
+        # the command trailer prints the return code (stdout) and the marker
+        # (stderr) on their own lines, so it always appends a newline to the real
+        # output; strip it to return the output verbatim (e.g. `printf abc` yields
+        # b"abc", not b"abc\n"). Under a tty the injected newline is translated to
+        # CRLF, so strip that instead of eating a real "\r".
         newline = b"\r\n" if self.isatty else b"\n"
         stdout_bytes = stdout_bytes.removesuffix(newline)
-        stderr_bytes = b"".join(stderr)
+        stderr_bytes = b"".join(stderr).removesuffix(b"\n")
         return stdout_bytes, stderr_bytes
 
 
@@ -341,7 +342,10 @@ class ShellSession:
             full_cmd = "true ; "
         full_cmd += f"printf '\\n%s\\n' \"$?\" ; echo '{marker}'"
         if not self.isatty:
-            full_cmd += f" ; echo '{marker}' 1>&2"
+            # the leading newline keeps the marker off the last line of stderr
+            # output, just like the return code on stdout; otherwise unterminated
+            # stderr output would hide the marker and hang communicate()
+            full_cmd += f" ; printf '\\n%s\\n' '{marker}' 1>&2"
         if self.custom_encoding:
             full_cmd_bytes = full_cmd.encode(self.custom_encoding)
         else:
--- a/plumbum/machines/session.py
+++ b/plumbum/machines/session.py
@@ -215,6 +215,12 @@ class SessionPopen(PopenAddons):
             self.returncode = "Unknown"  # type: ignore[assignment]
         self._done = True
         stdout_bytes = b"".join(stdout)
+        # the command trailer prints the return code on its own line, so it always
+        # appends a newline to the real output; strip it to return the output verbatim
+        # (e.g. `printf abc` yields b"abc", not b"abc\n"). Under a tty the injected
+        # newline is translated to CRLF, so strip that instead of eating a real "\r".
+        newline = b"\r\n" if self.isatty else b"\n"
+        stdout_bytes = stdout_bytes.removesuffix(newline)
         stderr_bytes = b"".join(stderr)
         return stdout_bytes, stderr_bytes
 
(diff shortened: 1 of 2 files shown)


henryiii:
I ran a Claude review, and it reported a finding:

I've analyzed the full diff and traced every code path in `session.py`. Let me present the review.

## Review of PR #830 — `fix(session): keep the return code off the last line of output`

**What it does:** Fixes #494/#275, where a shell-session command producing output with no trailing newline (e.g. `printf abc`) had the return-code echo glued onto the same line (`abc0`), breaking `int()` parsing so `returncode` became the string `"Unknown"` and output was corrupted/dropped. The trailer now prints the code on its own line via `printf '\n%s\n' "$?"`, and `communicate()` strips exactly the one injected trailing newline (CRLF-aware under a tty) so output is returned verbatim.

**Assessment:** The core fix is correct and well-reasoned. I verified it against many cases — no trailing newline, embedded-but-untrailed newlines, empty output, `true`, already-newline-terminated output, real trailing `\r`, and tty CRLF translation — and in every case the returned stdout matches the pre-fix behavior for newline-terminated commands while now being correct for untrailed ones. `bytes.removesuffix` is 3.9+ (fine for this project), and the stdout marker stays on its own line, so marker detection is unaffected.

### Findings

**`plumbum/machines/session.py:344` — the symmetric stderr-marker gluing remains and is a hang, not just a wrong return code (PLAUSIBLE, pre-existing).**
The non-tty stderr trailer `echo '{marker}' 1>&2` is unchanged. A command that writes to stderr without a trailing newline glues onto the marker line: `printf err 1>&2` yields `errmarker\n`, so `MarkedPipe.readline`'s `line.strip() == self.marker` check (`session.py:78`) never matches, the stderr source is never removed from `sources`, and `communicate()` blocks forever on the next `readline()`. Concretely, `local.session().run("printf err 1>&2", retcode=None)` hangs. This is worse than the stdout symptom the PR fixes (which only mis-parsed the code, never hung). It is pre-existing and the author explicitly scoped it out in the PR description — surfacing it only to confirm the maintainer's decision: the contained fix leaves a live hang on the stderr path, which is the argument for the deeper `MarkedPipe`/sentinel rework Henry suggested on #275.

Nothing else survived verification — the introduced code has no new correctness, cleanup, altitude, or CLAUDE.md-convention issues. The regression test is well-targeted (covers untrailed output, embedded newlines, and non-zero exit).

henryiii:

Pushed c18f1c5 addressing the review finding: the identical gluing on the **stderr** path is now fixed too, so this PR no longer leaves that behind.

The stderr case was worse than the stdout one — `echo '<marker>' 1>&2` glued the end-marker onto unterminated stderr output (`errmarker`), so `MarkedPipe.readline` never matched it and `communicate()` blocked forever. Reproduced: `local.session().run("printf err 1>&2", retcode=None)` hung until killed.

The fix mirrors the stdout approach: the trailer now prints the stderr marker via `printf '\n%s\n' '<marker>' 1>&2` so it always lands on its own line, and `communicate()` strips that one injected newline back off stderr. A plain `\n` suffices there since the separate stderr pipe only exists in non-tty mode. Same shared marker path, so local/`SshMachine`/`ParamikoMachine` are all covered.

Regression test extended with three stderr cases (unterminated, embedded newlines, newline-terminated stays verbatim) plus a 20s timeout marker, since the pre-fix failure mode is a hang, not an assertion failure — confirmed it times out exactly at the blocking `readline` without the fix. Full local suite green, ruff/mypy/pylint clean.

henryiii:
Let me know what you think of the added commit.

apoorva-01:
Looks good, thanks for taking it the rest of the way. The stderr path was the bit I'd left alone, and it's the nastier half: stdout just misparsed the code, here `communicate()` hangs forever.

Pulled the branch to check. On the parent commit `local.session().run("printf err 1>&2", retcode=None)` blocks past 20s like you said; with your commit it comes back clean and the session tests pass. Stripping a plain `\n` off stderr (vs the CRLF `newline` for stdout) is right too, since the separate stderr pipe only exists in non-tty mode.

Good to merge from my side, and nice that it kills the #275 stderr case too.

PR #834, opened 2026-07-21
fix(cli): don't change the global SIGPIPE disposition

henryiii:

Since 2.0.1 (#727), `Application.run()` set SIGPIPE to `SIG_DFL` process-wide. A user reported that this intermittently killed their rpyc servers on Fedora 43: rpyc's entry points are plumbum `Application`s, and with `SIG_DFL` a socket `send()` to a peer that already closed delivers a fatal SIGPIPE instead of raising `BrokenPipeError`. The Python docs warn against exactly this: https://docs.python.org/3/library/signal.html#note-on-sigpipe

This removes the signal change and instead catches `BrokenPipeError` around the body of `run()`, redirecting stdout to devnull before exit so the interpreter's final flush doesn't raise again — the pattern the docs recommend. The piped-help fix from #727 still works, and its test still passes.

Adds two regression tests: `run()` leaves the SIGPIPE disposition alone, and a `BrokenPipeError` from `main()` returns exit code 1. Both fail without the fix. Probably worth a 2.0.2 patch release.

### DIFF
--- a/plumbum/cli/application.py
+++ b/plumbum/cli/application.py
@@ -918,6 +918,46 @@ complete -F _{prog_name}_completion {prog_name}
 
         return out_args
 
+    def _parse_and_dispatch(self, argv: list[str]) -> tuple[Application, int]:
+        """Parses ``argv``, runs switches and ``main()``; returns the final
+        instance (the nested subcommand's, if one ran) and the return code."""
+        inst: Application = self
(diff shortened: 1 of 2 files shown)


pmachata:
Looking good, thanks for the fix!
Tested-by: Petr Machata <petrm@nvidia.com>

Issue #755, opened 2025-11-10
cli.ExistingFile: Scripts crash after upgrade to 1.10


tadada-o:
I use `pb's` `cli.ExistingFile` objects in my scripts to handle command line arguments of type "file".
This worked up to version 1.9  of  `pb`.

With version 1.10, all my scripts now fail with error message `'FileArgument' is not in list`.

The complete Traceback is
```
2025-11-09 11:04:58,418 - 💥️ - ERROR - extrahiere Barcodes aus PDF.py,569 - root.<module> - 'FileArgument' is not in list

Traceback (most recent call last):

  File "/Users/Guy/bin/extrahiere Barcodes aus PDF.py", line 565, in <module>

    ermittleDatenausBarcodesinPDF.run()

  File "/Users/Guy/Library/Python/3.12/lib/python/site-packages/plumbum/cli/application.py", line 608, in run

    ordered, tailargs = inst._validate_args(swfuncs, tailargs)

                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/Users/Guy/Library/Python/3.12/lib/python/site-packages/plumbum/cli/application.py", line 549, in _validate_args

    positional[args_names.index(item)] = annotation

               ^^^^^^^^^^^^^^^^^^^^^^

ValueError: 'FileArgument' is not in list
```

After reverting `pb`to 1.9, the error no longer occurs

workflowsguy:
After analyzing this issue with the help of a coding LLM, I now have at least some clue what is going on.

I use `cli.ExistingFile` in classes derived from a `FileApp` wrapper class like in the following code fragment

```
from pathlib import Path

from plumbum import cli
from autologging import logged, traced

from FileApp import FileApp, Color


@traced
@logged
class detectDateValues(FileApp):
    PROGNAME = "detect document creation date and orther relevant date values"

    def main(self, FileArgument: cli.ExistingFile) -> int:
        """
        This function contains the actual processing logic
        """
        self.File = Path(FileArgument)
        super().main(FileArgument, detectDateValues.PROGNAME)
        # initialize Logger for this class
        self.__log = logging.getLogger(self.__class__.__name__)
		# etc.
```

According to the LLM's analysis, the issue comes from a conflict with the `traced` and `logged`  decorators. They seem to change the class/function signatures somehow.
[...]

PR #838, opened 2026-07-25
fix(cli): ignore annotations for non-positional main parameters

Sanjays2402:
Helps with #755

An annotated keyword-only parameter on `main()` made `_validate_args` call `args_names.index(item)` for a name that is not a positional argument, raising `ValueError: 'verbose' is not in list` before the app could run (same shape as the reported `'FileArgument' is not in list` when a decorator reshapes the signature). Annotations are now only looked up when they name an actual positional argument.

Regression test in `tests/test_cli_annotations.py` fails with that `ValueError` without the change and passes with it.

### DIFF
--- a/plumbum/cli/application.py
+++ b/plumbum/cli/application.py
@@ -897,7 +897,7 @@ complete -F _{prog_name}_completion {prog_name}
             for item, annotation in annotations.items():
                 if item == m.varargs:
                     varargs = annotation
-                elif item != "return":
+                elif item in args_names:
                     positional[args_names.index(item)] = annotation
 
             new_tailargs = self._positional_validate(
--- a/tests/test_cli_annotations.py
+++ b/tests/test_cli_annotations.py
@@ -8,8 +8,20 @@ class App(Application):
         print(f"file={file.name}")
 
 
+class KeywordOnlyApp(Application):
+    def main(self, file: ExistingFile, *, verbose: bool = False):
+        print(f"file={file.name} verbose={verbose}")
+
+
 def test_access_annotations(capsys):
     _, rc = App.run(["prog", "pyproject.toml"], exit=False)
     assert rc == 0
     stdout, _ = capsys.readouterr()
     assert "file=pyproject.toml" in stdout
+
+
+def test_annotated_keyword_only_argument(capsys):
+    _, rc = KeywordOnlyApp.run(["prog", "pyproject.toml"], exit=False)
+    assert rc == 0
+    stdout, _ = capsys.readouterr()
+    assert "file=pyproject.toml verbose=False" in stdout


henryiii:
Thanks! I don't think this fully closes #755 (see review below if you want), so let keep that open for a potential follow-up.

<details><summary>:robot: Claude Opus 5 medium review :robot:</summary>

## Review: PR #838 — `fix(cli): ignore annotations for non-positional main parameters`

### Overview
One-line change in `plumbum/cli/application.py:900`: when building the positional-validator list from `main()`'s type hints, replace the `item != "return"` filter with `item in args_names`. `inspect.getfullargspec` puts keyword-only parameters in `kwonlyargs`, not `args`, but `typing.get_type_hints` returns them all — so any annotated keyword-only parameter caused `args_names.index(item)` to raise `ValueError` before the app could run. Plus a regression test.

### Verification
- Reverted the one-line fix locally: `tests/test_cli_annotations.py::test_annotated_keyword_only_argument` fails with `ValueError` at `application.py:901`. Restored: passes. The regression test does its job.
- Full suite: 483 passed; the 3 failures (`test_config.py::test_home`, `test_local.py::test_list_processes`, `test_pgrep`) are sandbox `PermissionError`s unrelated to this change.
- `prek -a --quiet` clean.

### Correctness
The fix is right and minimal. `item in args_names` subsumes the old `"return"` exclusion and additionally guards annotated `**kwargs`, an annotated `self`, and names that don't exist in the argspec at all. No behavior change for any signature that worked before.

### Main concern: #755 is silenced, not really fixed
The PR says "Closes #755" (the `'FileArgument' is not in list` case where a decorator reshapes the signature). That case is a *different* mismatch: `getfullargspec` does not follow `__wrapped__`, so it sees the wrapper's `(*args, **kwargs)`, while `functools.wraps` copies the inner function's `__annotations__`. With this change the crash disappears, but validation is silently dropped:

```python
class App(cli.Application):
    @deco                                   # functools.wraps wrapper
    def main(self, file: ExistingFile): ...
# now prints: type: str pyproject.toml     ← plain str, ExistingFile never applied
```

Arity checking is also lost, since the wrapper's `*args` makes `max_args` unbounded. Not a crash is better than a crash, but a user who wrote `ExistingFile` now gets no file check and no error. Suggestions, in order of preference:

1. Unwrap before introspecting — `inspect.getfullargspec(inspect.unwrap(self.main))` (or `inspect.signature(..., follow_wrapped=True)`) — so decorated mains keep their validators. That would genuinely close #755.
2. If that's too big for this PR, keep the fix as-is but retitle the PR to the keyword-only bug and leave #755 open, and add a test pinning the current decorator behavior so a future change is deliberate.

### Smaller points
- `plumbum/cli/application.py:896` — the comment `# All args are positional, so convert kargs to positional` is now stale and slightly contradicts the new filter. Worth updating to something like "only annotations naming a positional parameter become validators".
- Keyword-only parameters *without* a default still fail later with `TypeError` at the `self.main(*tailargs)` call, and `min_args`/`max_args` (lines 860–861) ignore `kwonlyargs` entirely. Out of scope here, but a reasonable follow-up — plumbum has no way to supply keyword-only values, so a clear error at class definition time would beat a `TypeError` at run time.
- Test coverage is adequate for the fixed bug. Two cheap additions: an annotated `**kwargs` on `main`, and a keyword-only parameter actually set via a `cli.SwitchAttr`/switch so the print asserts `verbose=True` on some path.

### Style, performance, security
Matches project conventions; test file layout mirrors the existing one. No performance impact (this path runs once per invocation, and `list.index` on an already-short list is unchanged). No security implications — except the note above, that a silently skipped `ExistingFile` validator means user input reaches `main()` unvalidated in the decorated case.

### Verdict
Approve the code change; it fixes a real crash with a correct, minimal edit and a proven regression test. Ask the author to either unwrap the signature or adjust the "Closes #755" claim, and to refresh the stale comment.

</details>

PR #841, opened 2026-07-25
fix: render bytes arguments as text instead of repr

### LINKED ISSUE #600: byte parameters and command names should be use as is
for example:

`print(plumbum.local['echo'](b'test'))`

should result in

`test`

but instead it results in:

`b'test'`

byte objects are not even supported in command names:

`plumbum.local[b'echo']`

results in a type error

`TypeError: a bytes-like object is required, not 'str'`

in `plumbum/machines/local.py:243`

Sanjays2402:
Closes #600

`local["echo"](b"test")` passed the `bytes` through `str()`, so the child process received the literal `b'test'` instead of `test`. Arguments now go through a small `_stringify` helper that decodes bytes with `os.fsdecode`, used by both `shquote` and the unquoted `formulate` paths.

Added `TestBytesArguments` in `tests/test_local.py` (fails on master, passes with the fix). Command *names* still require `str`; this covers arguments only.

### DIFF
--- a/plumbum/commands/base.py
+++ b/plumbum/commands/base.py
@@ -12,6 +12,7 @@ __lazy_modules__ = {
 
 import contextlib
 import functools
+import os
 import shlex
 import subprocess
 import typing
@@ -69,10 +70,20 @@ class RedirectionError(Exception):
 # ===================================================================================================
 # Utilities
 # ===================================================================================================
+def _stringify(value: Any) -> str:
+    """Renders a command-line argument as text.
+
+    ``bytes`` are decoded with :func:`os.fsdecode` rather than going through
+    ``str()``, which would render them as their ``repr`` (``b'test'``).
+    """
+    if isinstance(value, (bytes, bytearray, memoryview)):
+        return os.fsdecode(bytes(value))
+    return str(value)
+
+
 def shquote(text: Any) -> str:
     """Quotes the given text with shell escaping (assumes as syntax similar to ``sh``)"""
-    text = str(text)
-    return shlex.quote(text)
+    return shlex.quote(_stringify(text))
 
 
 def shquote_list(seq: Sequence[Any]) -> list[str]:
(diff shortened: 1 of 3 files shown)


henryiii:

Thanks! The core fix is correct — `local["echo"](b"test")` now sends `test`, and surrogate-escaped bytes round-trip through `os.fsdecode` → `subprocess` → `os.fsencode` on POSIX. I pushed a follow-up commit with three changes:

1. **`test_bytes_argument_in_list` did not test the list branch.** `BaseCommand.__getitem__` unpacks a `list`/`tuple` into `bound_command(*args)`, so `local["echo"][[b"a", b"b"]]` becomes two scalar arguments and never reaches the `isinstance(a, (list, tuple))` branch in `ConcreteCommand.formulate`. Reverting only that branch to `str(b)` kept the test green. The test now passes a nested list, which does reach the branch.

2. **The fix missed the "`args` is one object" path.** `LocalCommand.popen` and `BoundCommand.popen` special-cased only `isinstance(args, str)`. `bytes` is a sequence of `int`, so it was iterated: `local["echo"].run(b"ab")` gave `'97 98\n'`. Both now test `(str, bytes, bytearray)`.

3. **Documented the encoding limits.** `os.fsdecode` is locale- and platform-dependent. On Windows the filesystem codec is UTF-8 with `surrogatepass`, so `local["echo"](b"\xe9")` raises `UnicodeDecodeError` instead of running. The command's own `custom_encoding` is also ignored, so for `SshMachine`/`ParamikoMachine` the *local* encoding decides how a byte argument is rendered for the *remote* host. This is now stated in the `_stringify` docstring; making it encoding-aware needs more plumbing than this PR should carry.

One remaining item, left as-is: `StdinDataRedirection.formulate` now inlines decoded binary data into `echo {shquote(self.data)}`, where it previously showed a `repr`. `ShellSession.run` uses that string verbatim, so `(cmd << b"\x00\xff")` over a session can raise `UnicodeEncodeError` or write NUL bytes into the shell. Both renderings are wrong for binary data, and the `popen` path uses a temp file and is unaffected, so this is a separate issue.

PR #845, opened 2026-08-12
fix: parse ps output by header column offsets

### LINKED ISSUE #844: `list_processes()` misparses ps output when `ps` STAT is empty
`LocalMachine.list_processes()` assumes every `ps` output row contains a non-empty STAT field. It parses rows using whitespace splitting:

https://github.com/tomerfiliba/plumbum/blob/061efa49e2cd14ac35d87ed99584c62b42645c45/plumbum/machines/local.py#L429-L442

However, `ps` on Darwin can emit an empty STAT field. This occurs with any non-Apple built `ps` (`darwin.adv_cmds` build by nixpkgs in this case) because it lacks Apple's private process-information entitlement.

For example, an output row equivalent to:

```text
12345   501      python -m pytest
```

is parsed as:

```python
pid = "12345"
uid = "501"
stat = "python"
args = "-m pytest"
```

The empty STAT field is lost, and every subsequent field is shifted. Consequently, `pgrep("[pP]ython")` test fails to find the Python process. This appears to affect both the local and remote process-list parsers.

Refs: https://github.com/NixOS/nixpkgs/issues/551001, https://github.com/NixOS/nixpkgs/issues/374909

Sanjays2402:
Closes #844

`list_processes()` splits each `ps` row on whitespace, which assumes STAT is never empty. Some `ps` builds (e.g. the non-Apple Darwin build referenced in the issue) emit an empty STAT, so every later field shifted — the first word of the command line was read as STAT — and `pgrep()` stopped matching. Both the local and remote parsers now share a helper that locates the args column by its offset in the `ps` header line, so an empty STAT stays empty and args stays intact.

Regression test added in `tests/test_lib.py`; it fails against the old whitespace-splitting logic (`stat='python'`, `args='-m pytest'`) and passes with the fix. `tests/test_lib.py` and `tests/test_local.py` are green locally (138 passed, 3 skipped).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/plumbum/lib.py
+++ b/plumbum/lib.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
-__lazy_modules__ = {"inspect", "io"}
+__lazy_modules__ = {"inspect", "io", "re"}
 
 import inspect
 import os
+import re
 import sys
 from contextlib import contextmanager
 from io import StringIO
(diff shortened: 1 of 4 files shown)


henryiii:

Thanks! The empty-STAT case is a real bug and this is the right place to fix it, but locating the args column from the header alone is not safe: `ps` sizes each column to its widest *value* and does not widen the header to match. Any row whose PID, UID, or STAT overflows its header column shifts the args column to the right, and the fixed slice then cuts in the wrong place. Reproduced with procps in a container:

- UID `12345678` → `stat='Ss'`, `args='l python3 /t.py'`
- UID `2000000000` → `stat=''`, `args='Ssl python3 /t.py'`

The old whitespace split parsed both correctly. Large UIDs are common (AD/SSSD, LDAP, user namespaces; `nobody` is 4294967294), and the same applies to 7-digit PIDs and to STAT values of 6 characters or more, such as `S<Lsl+`. The `stat[0] if stat else ""` fallback also makes the failure silent: it reports the empty STAT that this PR fixes, and `pgrep("^python")` stops matching because `args` gets a `Ssl ` prefix.

I pushed a commit to the branch that keeps your fix and removes the assumption:

- Match the PID and UID on each row, then use the header args column only to decide if the field after the UID is a STAT value or the start of the command line. A field at or after that column (moved right by the row overflow) is the command line.
- Added a test with over-wide UID, over-wide STAT, and empty STAT rows.

